PR #36, opened 2022-08-24
Fix CPU usage on tiling WMs

0323pin:


woelper (approved):
Thanks. Tested on linux, too. Historically, I wanted to keep sync _off_ on linux, as this consumed more cpu with lazy. This does not seem to be the case any more!

PR #39, opened 2022-08-28
Add undocumented key shortcut

0323pin:


woelper (approved):
Thank you!

0323pin:
@woelper btw, the version on crates.io is still at 0.5.46

woelper:
Thanks for the reminder!
I just released both here and on crates.io. I also modified the release shell script so in the future the releases should always stay in sync between github and crates.io.

Have a great weekend!

Issue #93, opened 2022-12-14
Add app to Homebrew casks/MacPorts package managers to handle autoupdates


eugenesvk:

woelper:
This is an interesting topic. Would you be able to help out with this?
Are you aware oculante can auto-update through settings?

eugenesvk:
I can help with Windows (see https://github.com/ScoopInstaller/Extras/issues/10003)

Don't know anything about Ports, and also unlikely can help with Brew since it doesn't support Catalina anymore, so might be a challenge to test, but might test it later, hopefully should be easier for binary installs...

Yes, I'm aware of auto-update, it's nice to have as well, but then it's not always the best thing for the first launch experience to be having to remember updating an app (especially when it's manual)

eugenesvk:
Tried to add a cask (binary distributable) on Homebrew, but apparently casks on ARM **must** be signed and since your binary isn't it will not be accepted (doesn't pass the audit)
You might want to add your [own tap](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap), this is the relevant cask formula that I created before I learned about the signing requirement
https://github.com/eugenesvk/homebrew-cask/blob/oculante/Casks/oculante.rb

Also, not sure if the most annoying part — updating the tap version/sha on new releases — can be automated

eugenesvk:
Good news — I've managed to find a way for you to set up an automated workflow to update the homebrew cask on every new tag released on this main repo (there is a github action to open a pull request to update the version and another action to automerge it, so hopefully you won't have to do anything once it's set up)

You'd need to complete the following two steps:
- Copy this cask repo https://github.com/eugenesvk/homebrew-oculante and update the installation instructions with the new links (and then reference it in your workflow described below)
- Check out this [sample repo](https://github.com/eugenesvk/homebrew-bump) that would be the equivalent of your main repo https://github.com/woelper/oculante/, the ReadMe should have sufficiently described how to set up the autoupdates

(for example, if I push a tag in that `homebrew-bump` repo, then the `homebrew-oculante and` cask repo is fully autoupdated with the version taken from that tag, including automerging the PR)

woelper:
Thank you, this sounds amazing! I'll get on with that as soon as I fixed some of the issues I am working on.

Stoppedpuma:
[...]

PR #96, opened 2022-12-19
Update (un)install/update instructions

eugenesvk:
For Windows/macOS

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@
 *.log
 Info.plist
 paint/target
+
+# project's built/temp docs
+Changelog.html
+ReadMe.html
--- a/README.md
+++ b/README.md
@@ -27,14 +27,49 @@ Images may contain color information that is masked by the alpha channel. Althou
 ![Screenshot](res/premult.png "Screenshot")
 
 
-## Installation
-Just download the executable for your system from the releases tab (https://github.com/woelper/oculante/releases). No installation is required. In order to open images you can configure your system to open your desired image formats with oculante, drag them onto the executable or into the window. Right now the executables are roughly 10MB.
+## Installation, Update, and Uninstallation
 
-On NetBSD, a pre-compiled binary is available through the native package manager.
-To install it, simply run
-```sh
-pkgin install oculante
-```
+Manually
+
+  - Install: just download the executable for your system from the releases tab (https://github.com/woelper/oculante/releases). No installation is required. In order to open images you can configure your system to open your desired image formats with oculante, drag them onto the executable or into the window. Right now the executables are roughly 10MB.
+  - Update: use the app's builtin updater via ⚙`Open settings` → `Check for updates`
+  - Uninstall: just delete the executable from your system and the settings from
+    - Windows `~/AppData/Local/.oculante`
+    - macOS `~/Library/Saved Application State/com.github.woelper.oculante.savedState`
+
+Via a package manager
+
+  - __Windows__: a pre-compiled binary is available via [Scoop](https://scoop.sh/)
+    ```sh
+    # Install
+    scoop bucket add extras
+    scoop install oculante
+    # Update (might not be needed if you update all apps via 'scoop update *')
+    scoop update oculante
+    # Uninstall (except for settings at '~\AppData\Local\.oculante')
+    scoop uninstall oculante
+    ```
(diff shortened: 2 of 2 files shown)


Stoppedpuma:
Hi, sorry for the long wait, I'm going to close this in favour of #425. Thanks for the PR as I made some use of it!

eugenesvk:
Your instructions are incomplete: you have no package managers and not all uninstall paths are listed, would've been better if you added your edits on top of this PR instead of losing data like this

Stoppedpuma:
I'm slightly confused here, package managers and uninstall paths **are** listed? If you're talking about scoop and brew, this is because there was either no package or the package was abandoned (in the case of scoop it wasn't updated since 0.6.x if I recall).

image

eugenesvk:
You recall incorrectly, scoop is 0.8.23 and isn't abandoned, they have a mechanism to autoupdate an app

For brew there is also a package (am using it), though not linked in this PR as I though by the time of merging oculante would get its own copy, so didn't want to link to mine

Uninstall for Mac is missing ~/Library/Saved Application State/com.github.woelper.oculante.savedState

Stoppedpuma:
Your uninstall path is incorrect. This was confirmed directly by @woelper 
image

> For brew there is also a package

Is there a link you can provide? I'm unable to find this through either regular search, nor formulae search. If the package isn't available officially on brew, it's not going to happen until it's available there.

> You recall incorrectly, scoop is 0.8.23

You seem to be correct here, I probably found an outdated source. I'll add this soon.

eugenesvk:
> Your uninstall path is incorrect. This was confirmed directly by @woelper

It is correct and confirmed directly by looking at the created files. It's just incomplete as it doesn't cover user config, only app data, these are two separate folders

> If the package isn't available officially on brew, it's not going to happen until it's available there.

They you might have to wait forever since brew doesn't accept unsigned cask, and oculante isn't signed
But you don't have to wait at all, you can follow my guide from https://github.com/woelper/oculante/issues/93 to add the dev-official package and fix this repo as well (you could combine homebrew cask and the update workflow in just this repo based on [this](https://github.com/eugenesvk/homebrew-tap))

> Is there a link you can provide? 

You could use the link from the guide https://github.com/eugenesvk/homebrew-oculante (or the example above with a single repo), but it has the downside of requiring manual intervention since after 60 days of repo inactivity the actions stop working, so that's another resaons why creating woelper's own package is better - it will trigger on releases, not each X days, so wouldn't fail like this

Stoppedpuma:
> It is correct and confirmed directly by looking at the created files. It's just incomplete as it doesn't cover user config, only app data, these are two separate folders

You seem to have gotten to my comment prior to the edit, this directory has changed on master.

I'm going to say that for now adding brew instructions is going to be a no for now, while woelper of course can say otherwise and I'll follow his decision, the reason behind this is simply that there's no moderation process on your package, currently it's not my job, nor woelpers responsibility moderate an unofficial package. Additionally, your package seems to not be updated often as the last update prior to the most recent one was 10 months ago. If woelper wants to turn make / turn this into an official package then sure and I'll have this updated as soon as possible, but in the mean time this is going to be a no on my part.

@woelper Pinging you here to read over this in-case you want to override my decision or want to look into making the brew package official.

eugenesvk:
> Additionally, your package seems to not be updated often as the last update prior to the most recent one was 10 months ago.

That's incorrect, you're looking at a SAMPLE package which isn't supposed to be used directly, so it wasn't maintained. The other link is more frequently updated since that's the one I use. But also: as opposed to the update frequency of ... never? Which is better?

But more importantly, 

> the reason behind this is simply that there's no moderation process on your package

That's why I suggested creating the dev-official package and linked to a guide, and maybe you could even do that in this very repo (two separate repos aren't mandatory, could combine all workflows in one) since you're already a collaborator?

Stoppedpuma:
> That's why I suggested creating the dev-official package and linked to a guide, and maybe you could even do that in this very repo (two separate repos aren't mandatory, could combine all workflows in one) since you're already a collaborator?

The problem here is if it's going to be done, I would try to have it done correctly. In this case, correctly means in the official brew repos. I'll bring this up in conversation with woelper after the next release which will be out in a few days. If we can't get this into official repos, I'll see about creating a workflow.

Issue #101, opened 2022-12-29
Enhancement: Improved File/Folder Opening Options


Aligorith:
**Description:**
Currently, the options available for how to select a file / folder to open are a bit restrictive and/or buggy.

What works:
* Dragging an image onto an open instance
* Providing a path to a specific image as an argument to the program
  * **BUG:** If you provide the path of a folder instead though, it fails to load anything, and just sits there showing the loading message + the spinner icon
* Use left/right arrows (buttons on toolbar, or keyboard) to change between images within a folder

Additional methods that would be nice to have:
* Ctrl-O / F1 - Open a file browse dialog, that lets you select a folder from the file system to use
* Passing a folder will open the first file found, instead of hanging
* Using Ctrl-Scrollwheel can also be used to change between images
* Have an option to display a horizontal panel/bar across the bottom of the UI, where clicking / click-dragging anywhere in there will jump to another image in the folder. It should also have an indicator of where the current image sits within the folder.

Other minor issues noticed:
* The filenames should really be sorted using "[Natural Sorting](https://en.wikipedia.org/wiki/Natural_sort_order)", instead of the "character string" sorting that standard libraries typically give us. (e.g. Instead of getting something like `"slice_1", "slice_10", "slice_100", "slice_101", ..., "slice_11", ..., "slice_2", ...`, you get `"slice_1", "slice_2", ..., "slice_10", "slice_11", ..., "slice_100", "slice_101", ...` as expected)

**Example Implementations:**
Over last year's Christmas/New Year period, I ended up hacking support for a bunch of these into Emulsion (i.e. see [my fork with these features](https://github.com/Aligorith/emulsion)).

Back then, it was the closest to what I wanted from a lightweight image viewer. However, architectural issues ended up making it hard to move beyond those simple tweaks to trying to make it use egui for supporting many of the things you've added here (e.g. EXIF display + pixel-color sampling support). It now looks like it may be easier to try to bring over those customisations to this codebase instead / as well, giving me another image viewer option.

I'll see what I can do. If I get some of these working, I'll probably send a pull-request implementing those. Otherwise, hopefully some of these wishlist items show up in a future release :)

woelper:
All these are great points. Thank you for taking the time to write this up, and I am happy that this viewer comes already close to what you are looking for. I will take a look at your fork and start seeing what I can do. Please don't hesitate to open other tickets in the meantime!

Aligorith:
[...]

PR #104, opened 2022-12-30
101 filebrowse tweaks

Aligorith:
This implements the following changes (as per #101):
* Make it possible to pass a folder-path as a command-line arg, instead of requiring a file within that folder (to avoid having a loading message that won't go away)
* Ctrl-Scrollwheel can be used to go to next/prev images
* Next/Prev images order is sorted using Natural / Lexical Sort Order
* Next/Prev images tools cycle around after hitting the first/last image in the folder, instead of getting stuck
* Ctrl-O / F1 opens a native file dialog to allow the user to select an image to load

I've also taken the liberty to add these to a new section at the top of the changelog. Not sure if you usually auto-generate / update this or not. It was just helpful to keep track of what I'd already implemented.


woelper:
Thanks!
Yes, the changelog is auto-generated from conventional commits.

woelper:
This is a really cool addition, and I think we should merge that as it is. I am thinking later if would be nice to have some more timeline-like widget for stepping through folders or image sequences. I imagine something like reading all image paths into a Vec so you can scrub through it or jump to specific frame, what do you think? This would allow us to do more sanitizing/filtering in a cleaner way.

Aligorith:
Updated to remove the manual changelog changes that were causing merge conflicts.

src/main.rs
```diff
@@ -906,3 +950,29 @@ fn next_image(state: &mut OculanteState) {
...
 }
+
+// Show file browser to select image to load
+fn browse_for_image_path(state: &mut OculanteState) {
+    let start_directory =
+        if let Some(img_path) = &state.current_path {
+            img_path.clone()
+        }
+        else {
+            std::env::current_dir().unwrap()
```
woelper, line 961: This needs to be handled

woelper (approved):

PR #118, opened 2023-01-07
Update README to include new hotkeys introduced in #101/#104

Aligorith:
Update README.md to include the shortcuts introduced in #104


woelper (approved):

woelper:
Thanks!
I am currently automating this by running a build script or test to dump shortcuts, so some of the keys may look different n the future, just a heads up!

PR #144, opened 2023-02-22
fix: update jpegxl

Quackdoc:
minor update to jpegxl-rs decoder rework


woelper (approved):
Looks great, thank you!

Issue #162, opened 2023-03-29
Add oculante package to AUR


Adhalianna:
The motivation is similar to what was mentioned here in issue #93  but besides auto-updates managed by the package manager having oculante available in AUR would make it simpler for me to have it set-up automatically with other packages I consider core on my system (I just pipe the list of packages to pacman, the package manager, when I am starting with a fresh installation). There's also a chance more people would find out about oculante by searching for packages on AUR. 

I cannot promise help with this issue but if I find some spare time I will probably add some relevant resources here in the comments section.

woelper:
Thanks, and I am grateful for the feedback.
I am all up for building a package, but I think I need some support and instructions as I have not used arch/AUR. Is anyone allowed to add new packages there?

woelper:
There seems to be a cargo subcommand facilitating this. Could you try out this PKGBUILD please?
https://github.com/woelper/oculante/blob/master/PKGBUILD

woelper:
@Adhalianna I am happy to realease the package to AUR once you test and we figure out the best way this should be installed according to cumminity guidelines.

Adhalianna:
> @Adhalianna I am happy to realease the package to AUR once you test and we figure out the best way this should be installed according to cumminity guidelines.

In the next three days I should have some spare time finally and I'll try to look into it.

woelper:
I added a CI step for this. I am going to do a test release soon!

woelper:
Amazing, this seems to work!
https://aur.archlinux.org/packages/oculante

Thank you again for your awesome contribution, this was fun!

PR #169, opened 2023-04-02
PKGBUILD for publishing to AUR

Adhalianna:
This improves the pkgbuild suggested in #162 and adds a desktop entry for Oculante.

As is the PKGBUILD cannot be built with `makepkg` but that is _on purpose_. It tries to copy the desktop entry from the sources using the reference to the latest release but that file does not exist in any release _yet_. I have tested the script as well as I could with my system and I am quite confident that it will work fine. Commenting out the line that copies the desktop entry within the PKGBUILD causes it to work perfectly fine and this trick can be used temporarily to test it locally. It is quite similar to pkgbuilds of other Rust crates (e.g. [xplr](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=xplr-git), [eww](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=eww)) 

I went over [submission guidelines for AUR](https://wiki.archlinux.org/title/AUR_submission_guidelines) and as far as I can tell this format of pkgbuild is perfectly fine.

All that is left is to figure out how to publish it from CI. I have seen [some github actions for that](https://github.com/marketplace/actions/publish-aur-package) but I have no experience with github actions whatsoever and so I would rather touch them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .DS_Store
 *.log
 Info.plist
+PKGBUILD
 paint/target
 shortcuts.txt
 perf.*
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,17 +1,26 @@
 # Maintainer: Johann Woelper <woelper@gmail.com>
 pkgname=oculante
-pkgver=0.6.57
+pkgver=0.6.58
 pkgrel=1
-makedepends=('rust' 'cargo')
+depends=('aom' 'libwebp' 'expat' 'freetype2' 'gtk3' 'cairo')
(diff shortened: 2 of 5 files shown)


woelper:
Thanks, this looks great!
I will try to figure out how to integrate this into the release process. I'll look at the action as well. I guess it's mainly about importing the correct keys.

I am just reading up on AUR, but we could also think of having something like `oculante-git` which just installs the latest  git version (or even the latest from crates.io, like `cargo install oculante --force` does). In that case we only need to upload the `PKGBUILD` once I guess. Not sure if that is something that others do.

woelper:
Thanks!

PR #188, opened 2023-05-20
feat: Native rust JXL support

woelper:


src/utils.rs
```diff
@@ -752,16 +721,48 @@ pub fn open_image(img_location: &Path) -> Result<FrameCollection> {
...
+                let result = renderer
+                    .render_next_frame()
+                    .map_err(|e| anyhow!("{e}"))
+                    .context("Can't render JXL")?;
+                match result {
+                    RenderResult::Done(render) => {
+                        let i = render.image();
+
+                        // JXL with alpha
+                        if render.extra_channels().len() == 1 {
```
EugeneVert, line 738: Thank you for integrating jxl-oxide
```suggestion
                        if renderer.pixel_format().has_alpha() {
```
Because there are many types of extra_channels, such as alpha, depth, spot color, selection mask, black, thermal, etc.

woelper:
Thanks, good point. I will probably invest a little more time, as there can be images with grey and alpha etc - doing some more sanity checks!

src/utils.rs
```diff
@@ -752,16 +725,117 @@ pub fn open_image(img_location: &Path) -> Result<FrameCollection> {
...
+                        debug!("{:?}", renderer.pixel_format());
+                        match renderer.pixel_format() {
+                            PixelFormat::Graya => {
+                                let float_image = GrayAlphaImage::from_raw(
+                                    framebuffer.width() as u32,
+                                    framebuffer.height() as u32,
+                                    framebuffer
+                                        .buf()
+                                        .par_iter()
+                                        .map(|x| x * 255.)
```
EugeneVert, line 771:
```suggestion
                                        .map(|x| x * 255. + 0.5)
```

EugeneVert: When quantizing to 8-bit, it is better to use `round()`, not `floor()`

src/utils.rs
```diff
@@ -752,16 +725,117 @@ pub fn open_image(img_location: &Path) -> Result<FrameCollection> {
...
+                                image_result = DynamicImage::ImageLumaA8(float_image);
+                            }
+                            PixelFormat::Gray => {
+                                let float_image = image::GrayImage::from_raw(
+                                    framebuffer.width() as u32,
+                                    framebuffer.height() as u32,
+                                    framebuffer
+                                        .buf()
+                                        .par_iter()
+                                        .map(|x| x * 255.)
```
EugeneVert, line 785: Also here

src/utils.rs
```diff
@@ -752,16 +725,117 @@ pub fn open_image(img_location: &Path) -> Result<FrameCollection> {
...
+                                image_result = DynamicImage::ImageLuma8(float_image);
+                            }
+                            PixelFormat::Rgba => {
+                                let float_image = RgbaImage::from_raw(
+                                    framebuffer.width() as u32,
+                                    framebuffer.height() as u32,
+                                    framebuffer
+                                        .buf()
+                                        .par_iter()
+                                        .map(|x| x * 255.)
```
EugeneVert, line 799: And here

src/utils.rs
```diff
@@ -752,16 +725,117 @@ pub fn open_image(img_location: &Path) -> Result<FrameCollection> {
...
+                                image_result = DynamicImage::ImageRgba8(float_image);
+                            }
+                            PixelFormat::Rgb => {
+                                let float_image = RgbImage::from_raw(
+                                    framebuffer.width() as u32,
+                                    framebuffer.height() as u32,
+                                    framebuffer
+                                        .buf()
+                                        .par_iter()
+                                        .map(|x| x * 255.)
```
EugeneVert, line 813: And here

PR #201, opened 2023-06-21
restore window position on startup, add support for notan 0.10.0

jonatino:
Added support for passing window position in https://github.com/Nazariglez/notan/pull/272 so that it happens during creation and avoids the window moving around on screen after it already loads.

Don't merge this until notan 0.10.0 is released.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,6 +45,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "android-activity"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


woelper:
Thanks so much for the PR!

Nazariglez:
Added fix for darker colors here: https://github.com/jonatino/oculante/pull/1

woelper:
Hi! Thanks again for the PR!
I discovered some other issues (for example the order for scaling/translation matters now) and color issues which I fixed in https://github.com/woelper/oculante/pull/223. I hope you don't mind I am closing this one. This was really helpful!

PR #223, opened 2023-09-14
Notan 0.10

woelper:


jonatino:
@woelper just tried the latest build of oculante but window position is no longer restored.

I can't see where we set the windows x/y position based off the saved location in your updated changes:

https://github.com/woelper/oculante/blob/master/src/main.rs#L121

I think we also need to set the windows x/y as well as the width + height here.

woelper:
Thanks for the comment!

I think the size was saved, but never the position. The challenge with this is:

On every platform, we need to make sure we have covered edge cases, for example:

- I move the window from my HD laptop screen to my 4k external one, close oculante, open with no external monitor again
- move window to an external one as above, close, open again (with the small one being main screen)

Since this needs a bit of thinking, probably cross-platform multi-monitor support and testing on all operating systems and window managers I am unsure the benefits outweigh the risks of this feature to be honest...

hUwUtao:
how about adding a hardcoded min size? sometimes, somehow it turn into a 0x0 window which render on screen as a literal 2px bar on windows
[src/main.rs:124](https://github.com/woelper/oculante/blob/52c719e4aff80ff3545d0582707bc981df826e1e/src/main.rs#L124)

woelper:
Very good idea. I'll create an issue for that.

PR #227, opened 2023-09-19
Fix: remove LTO from pkgbuild

woelper:


res/pkgbuild
```diff
@@ -10,6 +10,7 @@ url="https://github.com/woelper/oculante"
 source=("$pkgname-$pkgver.tar.gz::https://github.com/woelper/${pkgname}/archive/refs/tags/${pkgver}.tar.gz")
 sha512sums=('SKIP')
 license=('MIT')
+options=('!lto')
```
LuisB79, line 13: it works!, this is for arch packages who have lto enabled by default

LuisB79 (approved):

Issue #255, opened 2023-12-20
Image is occluded by side panels


B-LechCode:
An image is partially occluded by the side panels, when fit to the window.
This is annoying when dealing with ultra wide images, in my case: 16384x1024 pixels.

I wanted to work out a fix. Would this be useful, or is the current behavior intended?

A question for that: where may I query the width of the two side panels? Is it currently tracked somewhere, or may I add their widths to the state struct?

Thanks :)

B-LechCode:
I'm working on this. Will provide a PR when I'm done with it.

woelper:
Hi! Thanks!
Layouts in egui are dynamic and are recalculated per frame. I would not store that somewhere but instead query egui for the sizes. As this is per-frame, I think that would have to be done at a place where the image is reset / received.

woelper:
Hi, and thank you again for your work!

I think your feature request is super good - I came to the conclusion that this should be the default how sizing is handled. With that in mind I made a new PR that directly modifies the function to scale - I hope this keeps the functionality a bit simpler while still keeping the original idea:

https://github.com/woelper/oculante/pull/267

Would you mind going for this simpler approach per default?

B-LechCode:
It's perfect to use that approach, I just kept the old approach because I thought it was intentionally.

But you omitted some of the other small fixes I included with my PR:
- Centering and scaling the image to the available draw area, you use the whole window height for centering/scaling: If an image has a high aspect, you will get the same problem I had when I reported that issue. Suggestion: please copy the calculation part I provided.
- Truncating path strings that are too long to keep the left panel at a steady width
Both are positive aspects for the look and feel of the gui in my opinion.

PS: It would have been ok, if you would just have edited my branch as you should have had the rights to do so :) Maybe it would have saved you some time.

woelper:
Thanks for the answer!

> Centering and scaling the image to the available draw area, you use the whole window height for centering/scaling: If an image has a high aspect, you will get the same problem I had when I reported that issue. Suggestion: please copy the calculation part I provided.

Are you referring to the menu bar occluding the image? I've tested lots of images and none occluded the side panels.

> Truncating path strings that are too long to keep the left panel at a steady width

I was not able to replicate that with a super long string. Also the left panel could potentially resize again in the future, so the approach should take a dynamic width into account. If you have a specific use case that fails I'd be happy to test it!

> PS: It would have been ok, if you would just have edited my branch as you should have had the rights to do so :) Maybe it would have saved you some time.

[...]

PR #260, opened 2023-12-24
Image bounds

B-LechCode:
Implemented a possible fix for #255 

Default mode is preserved as is, I added a new mode for the image geometry, which limits the area for the image to the available area.

Had to reduce the maximum file name char count from 20 to 18, otherwise the left panel unintentionally changes with by varying file name lengths.


B-LechCode:
Hi @woelper,

I discovered a little problem: the width of the path string is sometimes too wide for the panel width.
Do you know a convenient method to estimate the strings rendered width?
I think, going for a uniform width string isn't a good idea from a design point of view, is it?

Thank you :)

woelper:
I have a couple of thoughts, but no good solution either:
Egui provides an overflow character which you can try at https://www.egui.rs/ in the "Text Layout" demo.
I think calling truncate (https://docs.rs/egui/latest/egui/widgets/struct.Label.html#method.truncate) on a label could be a way to start.

My suggestion would be to let egui handle the layout completely and get the width of the available size with `ctx.available_rect()` after the panels have been added for example. I have not tried it, but the docs look like this may be the only info you need perhaps:
https://docs.rs/egui/latest/egui/struct.Context.html#method.available_rect

B-LechCode:
Using available_rect for calculation of the image dimensions is what I did, but only once after the image is reset, to allow dragging and zooming. In the case of an overflowing text, the left panel expands afterwards. A possible fix could be: calculating this twice, but in my opinion it's not good either, it's better to directly target the problem's source.
Using truncate would be possible, but in my opinion somehow ugly, because of truncating the string on "both sides". What do you think about that?

I found this, but didn't understand how to implement yet: https://github.com/emilk/egui/discussions/1488
If this would work, I could iterate the "front" string truncation until it fits. Having a fixed width panel when browsing images is the best approach, I think.

woelper:
> Using available_rect for calculation of the image dimensions is what I did, but only once after the image is reset, to allow dragging and zooming. In the case of an overflowing text, the left panel expands afterwards, while rendering, I think. A possible fix could be: calculating this twice, but in my opinion it's not good either, it's better to directly target the problem's source. Using truncate would be possible, but in my opinion somehow ugly, because of truncating the string on "both sides". What do you think about that?
> 
> I found this, but didn't understand how to implement yet: [emilk/egui#1488](https://github.com/emilk/egui/discussions/1488) If this would work, I could iterate the "front" string truncation until it fits. Having a fixed width panel when browsing images is the best approach, I think.

I _think_ they talk about layouting the text in a separate step before painting it, and then using the layout information to get the length.

I am not sure I understand what you mean by "truncating on both sides". Would it not be possible to omit the "rightmost" part of the filename?

B-LechCode:
It is indeed possible to omit the rightmost part of the filename, but I don't think it's pretty, truncating the leftmost part only seems better. I implemented rightmost truncation it in my last commit, please have a look :)

woelper:
Closing this as addressed by https://github.com/woelper/oculante/pull/267 - thanks a bunch!

PR #274, opened 2024-02-11
Truncating path label width for consistent left panel width

B-LechCode:
A simple fix for the last small inconvenience described in #255


woelper (approved):

woelper:
Thank you so much!

PR #283, opened 2024-02-28
parse uri paths given in input

diegostafa:
currently oculante can't read an input in a URI format (file:///this/is/a/valid/file)

i've added fluent_uri as a dependency to extract the path before mapping it into a PathBuf


woelper:
Thanks!
Could this also be relevant in other scenarios, such as a dropped file?

diegostafa:
i can't find a program exposing a uri when dragging a file, i don't know if Event::Drop(file) would even fire in that case

woelper (approved):

woelper:
Thank you!

woelper:
Unfortunately this broke the ability to open files on Windows. I disabled this on windows:
https://github.com/woelper/oculante/pull/295

PR #291, opened 2024-03-06
Fix high cpu usage when started without images

diegostafa:
main_menu was requesting a frame for the loading spinner indefinitely


woelper (approved):
Thanks!

PR #303, opened 2024-04-01
Add mouse wheel setting

ykszk:
Currently, mouse wheel and ctrl-wheel are statically assigned to zooming in/out and moving to next/prev image respectively.
With this PR, users will be able to swap these assignments.

## Changes
- add a new persistent setting `mouse_wheel_action`
- add `Mouse` section in the preferences UI for changing the setting (screenshot below)

mouse_wheel_setting

## New behavior
- With `Zoom` selected (default), mouse wheel works as zoom in/out and ctrl-wheel works as next/prev.
- With `Next/Prev` selected, mouse wheel works as next/prev and ctrl-wheel works as zoom in/out.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,6 +14,7 @@ use notan::draw::*;
 use notan::egui::{self, *};
 use notan::prelude::*;
 use shortcuts::key_pressed;
+use shortcuts::MouseWheelEvent;
 use std::io::Read;
 use std::path::PathBuf;
 use std::sync::mpsc;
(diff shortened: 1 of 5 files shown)


woelper:
Sorry for the delay!

First of all, thank you for this PR!

I really like this feature. I am thinking if we could solve this in a more generic way - for example could we maybe just expose wheel up and wheel down and have users set his directly, by setting shift + scroll to a specific action? What do you think?

Stoppedpuma:
Hello @ykszk! Sorry for the long delay but I am going to close this PR as it's heavily outdated and is vastly different both visually and for how keybinds will be handled going forward. This will take time to be implemented as this requires the preferences redesign to be implemented. Thanks for the work you put into this PR! An issue related to this is #408.

PR #321, opened 2024-05-10
Big images

B-LechCode:
@woelper: I finally implemented the tiling feature for the new approach.
Rendering the preview image now seems to be stable.
There's one downside: when the preview render consists of multiple textures, there's a small gap between the textures. I don't know why. I'd like to draw that via notan, too as this didn't have this problem for rendering the main image. But I don't know how to add a "draw" to the render pipeline at that point.

Could you help me with that, please?

After making it all work, I will refactor this and do some safety stuff - after that I will bring up a PR for master :)


woelper:
Sorry, I have not forgotten this. It is just hard to find the time this deserves!

I hope I can help with the info window, but I've seen that egui renders the texture differently than notan. I am unsure if we can reuse the rendering code there as well, maybe this requires some creative problem solving! :)

B-LechCode:
> Sorry, I have not forgotten this. It is just hard to find the time this deserves!
> 
> I hope I can help with the info window, but I've seen that egui renders the texture differently than notan. I am unsure if we can reuse the rendering code there as well, maybe this requires some creative problem solving! :)

I chatted back to my boss, I will try to get an interesing example image out of our line scan system.
The latest commits are at least "okay" from that perspective - there is a small margin between tiles, but if you imagine textures of 16k by 16k, they won't show up frequently.
Please have a look - if you're okay with it, I will start cleaning the code up. :)

B-LechCode:
Hi @woelper, 
did you have time to have a look at this since then?

woelper (approved):

PR #357, opened 2024-06-09
Compare improvements

woelper:


Stoppedpuma:
@woelper Small oversight, `ClearImage` isn't registered and doesn't appear in the shortcuts list.

`ERROR oculante::shortcuts] Command not registered: 'ClearImage'. Inserting new.`

woelper:
The error itself is expected, this happens if your configuration does not have a value that has been added to Oculante. If it is missing in the shortcuts list, this of course sounds like a bug!

Stoppedpuma:
Yep, missing in the shortcut list.

woelper:
This is fixed in master now.

Stoppedpuma:
@woelper This now causes all images to clear instead of just the current one.

PR #368, opened 2024-06-16
Add cargo feature recent_images

pauschuu:
If you want to build a more privacy-focused version of the App you can now remove the cargo feature "recent_images" so they won't be stored in the config file.


woelper:
I like the idea - but should we not make that a menu option instead of a cargo feature? If someone wants privacy, they could just toggle a flag instead of building the app with a parameter.

pauschuu:
> should we not make that a menu option instead
I will have a look into it this week hopefully. Closing this PR.

PR #417, opened 2024-09-08
truncate float offset to align to pixel coords

exhu:
The changes fix texturing artifacts at 1:1 scale for images which are displayed with a non-whole offset when window size or image size is odd. Added an example image to the tests directory which reproduces the problem. The test image is odd sized, and has the last column and row filled with transparent pixels to test the checkers background.


woelper (approved):
LGTM, thanks!

PR #422, opened 2024-09-15
feat: Stack Blur provides much faster blur performance for the blur f…

woelper:
This replaces `imageops::blur` with `libblur`'s Stack blur, which is a O(1) operation and works much faster in small and especially large images.

### DIFF
--- a/src/image_editing.rs
+++ b/src/image_editing.rs
@@ -384,7 +384,7 @@ impl ImageOperation {
                 }
                 r
             }
-            Self::Blur(val) => ui.slider_styled(val, 0..=20),
+            Self::Blur(val) => ui.slider_styled(val, 0..=254),
             Self::Noise { amt, mono } => {
                 let mut r = ui.slider_styled(amt, 0..=100);
                 if ui.checkbox(mono, "Grey").changed() {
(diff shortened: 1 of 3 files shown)


Stoppedpuma:
Extremely impressive performance compared to the previous solution! I'm no longer experiencing any freezes (it's closer to the program dropping to single frames per second rather than a freeze) on a 6000x6000 image. 

While performance is good, this PR does seem to break moving filters up and down.

src/image_editing.rs
```diff
@@ -384,7 +384,7 @@ impl ImageOperation {
                 }
                 r
             }
-            Self::Blur(val) => ui.slider_styled(val, 0..=20),
+            Self::Blur(val) => ui.slider_styled(val, 0..=254),
```
Stoppedpuma, line 387: Any reason 254 instead of 255/256? 255 is a more common number found and 256 of course follows bit increments.

woelper: Yeah, it is a limitation of the filter:
https://docs.rs/libblur/latest/libblur/fn.stack_blur.html

`radius - radius is limited into 2..254`

woelper:
> While performance is good, this PR does seem to break moving filters up and down.

I suspect something else is happening here: There are two different filter types: Per pixel and per image. For performance reasons you can only swap between the same types. 

Per image filters are

```
Blur
Resize
Crop
CropPerspective
Rotate
Flip
ChromaticAberration
LUT
Filter3x3
ScaleImageMinMax
```
The rest are per pixel filters. You can reorder between the same types. Maybe we could somehow visualize this better.

Stoppedpuma:
> The rest are per pixel filters. You can reorder between the same types. Maybe we could somehow visualize this better.

I'll think of a better way to sort this in the redesign, a temporary solution for now could be having the background colour be slightly different for the two types, similar to the alternating background colour currently present.

woelper:
That would be good. I am trying out a separator between them and greying out the up/down arrows in the meantime.

As you can guess, the whole image editing portion hasn't seen much love :)

PR #428, opened 2024-09-17
Recent menu redesign

woelper:
This PR merges the current state of the menu redesign. While not completed fully, some features provide significant UI/UX improvements.


Stoppedpuma:
@woelper One major issue I forgot to mention was spaces seem to be doubled
image
.

woelper:
> @woelper One major issue I forgot to mention was spaces in text seem to be doubled image .

Maybe we can leave that for the next iteration? Do you think this is a critical issue?

Stoppedpuma:
Depends, if it's a two minute fix I think it's worth fixing prior to merge as it's pretty visually distracting, otherwise if it will take a while then the next iteration is fine.

Stoppedpuma:
The title bar being huge is something which needs to be fixed.
image

Stoppedpuma:
The colour histogram seems to have broken, by default it's extremely zoomed in, not centred, and is either too high, too low, or too far to the left or right.
image
image

Stoppedpuma:
These don't seem to match how far to the right the buttons / other text are.
image

Stoppedpuma:
@woelper These two should be the last things to be fixed and this should be good to merge! First, the spacing for the icons on the file picker "breadcrumbs" seems to have regressed. Secondly, the previous directory button is slightly too short, 3px when I counted.

image

Stoppedpuma (approved):
Should be ready to merge after fixing the icon spacing on the "breadcrumbs" bar in the file manager. The panels are also the incorrect colour again (#1b1b1b) instead of #191919 (25, 25, 25) in rgb terms but this is rather minor and can wait until the next revision.

image

PR #452, opened 2024-10-06
Big images

woelper:


woelper:
@B-LechCode Sorry for all the delay and a BIG THANKS for your work. I have added my latest changes from master to do EXIF rotation and will merge this as soon as the checks pass.

PR #479, opened 2024-10-24
Folder bookmarks and save options

woelper:


Stoppedpuma:
The save and format buttons seem to be offset quite high. The formats name will also duplicate if it's not set to jpg.

image

Stoppedpuma:
Looks good to me besides a few colours being incorrect, but this can wait for now if you want to get this in asap.

woelper:
Love that you spotted that. Will fix!

PR #489, opened 2024-10-27
wip: try alternate zoom option without egui

woelper:


B-LechCode:
@woelper I added the exact coordinates to render the notan stuff.
I think this is an excellent alternative to egui::image, because it supports more texture types and we can use the same shaders as we use on the main image.

I'll try to improve, maybe we should merge this to dynamic image then, because it's tightly related.

woelper:
Thank you! Yes, I think you are right - it makes 100% sense now to merge this to dynamic images.

PR #493, opened 2024-10-28
Fixing: 486 all options under alpha tools cause crash

B-LechCode:


woelper:
I have an another idea: How about instead of modifying the image we send the updated image the same way we do when loading an image? For example for the alpha tools, we process the image, then send the buffer? It will be automatically displayed next frame.

woelper:
Hey! I put up a PR here:
https://github.com/woelper/oculante/pull/494

I think I'd like to go more in that direction as we don't add more to the state and add more checks. If we just send another image it will lead to updating the texture centrally. If this is not a performance issue, we could also do that in more places. One advantage would be that heavy image processing could be done in a thread and just sent like any regular image and then displayed as it comes in.

B-LechCode:
Closed in favor to the better solution @woelper worked out.

PR #494, opened 2024-10-28
fix: Send frame wen editing alpha tools to prevent crash

woelper:


Stoppedpuma:
Just a heads up performance wise, this causes the UI to freeze for ~800ms on a 16k image. The delay isn't as noticeable at on smaller resolutions such as 8k (takes around 200ms) but I also have a high end system so this could still be bad for those on lower end hardware.

woelper:
Thanks for checking! This is a bit odd, as the same function is called. Then the sending is probably to blame. Were you testing in release mode?

Stoppedpuma:
This seems to happen with both `cargo run` and `cargo build`, same applies with manually specifying to use the release profile.

woelper:
I've run an older version of oculante, and I get this behavior too. I've removed an unnecessary `clone()` which might speed things up, but I don't think the delay comes from the change. I've tested to only run the actual function without anything else (`highlight_bleed(img)` for example) and get the same delay.

Stoppedpuma:
Seems I've just never noticed this before, thanks for testing!

woelper:
I pushed the clone removal, so it should be faster. I experimented with doing this in a thread - it is instant, but I tested it on an 8k image and could not really tell the difference on my 3y old laptop - I don't think it is worth the complexity to be honest.

Stoppedpuma:
It's definitely faster, thanks!

B-LechCode:
@woelper interesting solution, you prefer this over the one I brought up, right? The first thing I see is that you always need to clone the image for sending. At is point we need 3 times the memory and have a performance drawback in comparison to my solution. Maybe I'm getting it wrong, but is there an option for moving instead of cloning?
I didn't have had time yet to compile it, and I need to think about profiling this. From the code standpoint, your solution is more beautiful, than mine :)
@Stoppedpuma Did you maybe compare the performance of this with my solution for the problem?

woelper:
Yeah, I think your solution is much faster! I am a bit manic at this point about keeping complexity down, so I think for this case we should go with an easier approach. It is only for a very rarely used function, and the memory should get freed immediately. It also only runs once on button press.
As for speed, I don't think there is a big difference, except for the reset part, but this can be improved (see your idea below)

In other areas this approach would be worse, for example where it needs to run frequently. We could send a special Frame without data to force texture updates centrally as you suggest. If the frame arrives, the texture will be updated. As you say, it should work with the cached edits - great idea!

This could potentially remove all the update code from the edit UI... sounds like something we could try!

And sorry for creating this PR out of the blue, I had no intention of stomping on yours. I just wanted to mess around a bit and try different solutions!

Stoppedpuma:
> I am a bit manic at this point about keeping complexity down

I don't necessarily think keeping minimal complexity at all times is the best way to handle things if it means losing out on medium to large performance increases. I think as long as things are kept tidy and aren't going to be 3 simultaneous migraines to maintain then a bit of complexity is alright.

> @Stoppedpuma Did you maybe compare the performance of this with my solution for the problem?

I previously did not, after testing your solution though it is slower than woelpers after the additional clone removal, prior to that commit they are relatively the same **with cargo run**, in a build yours is faster I think? I don't have time at the moment to do any profiling but I will post some results sometime tomorrow.

woelper:
I've now added the option to send an empty "marker" frame. This avoids having to clone the image. When an updated image needs to be sent, I think the performance does not suffer too much as the same functions are called as before.

B-LechCode:
@woelper Thanks for updating :)
What would you think of changing the empty frame to some kind of enum - maybe by extending the FrameSource enum?
We could then directly calculate and set alpha bleed etc. without the need of pushing the data through the channel? Wouldn't this save us one "copy operation"? 
This might be wrong, I'm not sophisticated enough to know.

Nevertheless: I like your solution :)

woelper:
> @woelper Thanks for updating :) What would you think of changing the empty frame to some kind of enum - maybe by extending the FrameSource enum? We could then directly calculate and set alpha bleed etc. without the need of pushing the data through the channel? Wouldn't this save us one "copy operation"? This might be wrong, I'm not sophisticated enough to know.
> 
> Nevertheless: I like your solution :)

You know what? I thought about enums as well while I was refactoring, but I was just too tired and wanted to finish it. I'll revisit it and try an enum approach. As you say, we could extend it with more functionality. Thanks for your feedback, always appreciated!

B-LechCode (approved):

B-LechCode:
At this point I think it can be merged.
Having more control via an enum would be just the the cherry to the pie.

Tested it and it works just fine. Working with a 8k by 32k image is slow, but that seems to originate from texture generation/updating. Maybe I should profile this sometime.

woelper:
@Stoppedpuma and @B-LechCode , I ended up refactoring everything to use a `Frame` enum. This gets rid of one struct and fixes an issue with compare mode, which seemed to be broken on master as well.

Could I ask you for a favor and build this branch and test a bit? I am afraid I broke something, although I've done lots of texting already.

Stoppedpuma:
List of issues I've found so far in my testing:

Reset all edits loses the images

Fullscreen and unfullscreening an image in the compare menu allows you to update position with a default position one (which is fine), the issue here is you can refullscreen and unfullscreen again and it has an update position option.

Adding two images to compare mode and drawing on one of them, then switching to the other and trying to draw on it switches back to the first image.

woelper:
> List of issues I've found so far in my testing:
> 
> Reset all edits loses the images

Nice. I fixed this!

> 
> Fullscreen and unfullscreening an image in the compare menu allows you to update position with a default position one (which is fine), the issue here is you can refullscreen and unfullscreen again and it has an update position option.
> 

I think this is an unrelated bug. It should of course be fixed, but I think this may be out of scope. Maybe we should take this and the point below to improve the compare mode further.

> Adding two images to compare mode and drawing on one of them, then switching to the other and trying to draw on it switches back to the first image and the second one still stays selected. Update: this seems to happen with all filters.

Great point, I think this is a generic problem. We can "fix" this by hiding the edit menu in compare mode maybe?

> draw_issue.webm

PR #495, opened 2024-10-28
Dynamicimage

woelper:


B-LechCode:
@woelper I started looking in that and tried to render grayscale textures - which didn't work. I created them as R8, but I can't tell notan to add a swizzle mask or something else to render it grayscale.

Doing "imageops::crop_imm" on a "DynamicImage" will always result in an rgba image, even if the underlying data is grayscale.

I didn't find any flaws in your implementation so far.

woelper:
Thanks so much for looking at it! Regarding grey textures, that is of course bad news. Maybe we can use a shader to convert red to grey? Would it make sense to open an issue in `notan` for this?

B-LechCode:
I thought about asking in notan's issue section, too.
I will do this rn :)

Yes, we maybe could define a default shader, doing conditional conversion?

woelper:
Great!

Yeah, a shader could make sense anyways for HDR and other things!

B-LechCode:
@woelper do you know where to inject a pixel shader to do that? That's beyond my knowledge. Maybe you got some hints.

Then I would try to commit a dynamic rendering for R8 at least.

woelper:
I was going to post the same URL :)

I am quite happy with the progress, your idea to have a convert operator was amazing. The pixel operators and image operators already support some additional data types and you can convert on the fly between them.

B-LechCode:
I got it for image rendering, but not for egui::image.
Maybe you have an idea where to add a shader for egui.
The alternative is maybe to use notan here, too. (Need to look at it today!)

Screenshot_20241102_004043

src/texture_wrapper.rs
```diff
@@ -198,15 +272,42 @@ impl TexWrap {
...
-                    gfx,
-                    my_img.as_ref(),
-                    my_img.width(),
-                    my_img.height(),
-                    settings,
-                    allow_mipmap,
-                );
+
+                let tex: Option<Texture>;
+                match image.color() {
```
woelper, line 277: I think you can match `image` directly here. This would prevent you from having to unwrap() inside the block

B-LechCode: Done!

B-LechCode:
@woelper is the branach dynamicimage_zoom supposed to combine the new zoom image and dynamic image?
I'd merge them then together and delete both old branches?

woelper:
Hi! I tried to combine them by merging zoom into it and made it build, but I think I am missing something. If you have time maybe you can try integrating the changes from your zoom window branch directly into this one? I can delete the `dynamicimage_zoom` in that case.

B-LechCode:
> Hi! I tried to combine them by merging zoom into it and made it build, but I think I am missing something. If you have time maybe you can try integrating the changes from your zoom window branch directly into this one? I can delete the `dynamicimage_zoom` in that case.

Done :)

woelper:
Amazing!!

B-LechCode:
> Amazing!!

Working on adapting zoom image to Notan rendering.
That might take some time. Will come back when done.
Adding the shader here, too works. So I think we can go on :)

Opening really big grayscale images got extremely faster!

B-LechCode:
@woelper
There are some types in Dynamicimage which don't have corresponding types in notan. All La-Types for example.

I thought about using a texture n times wider (double the width for La images) and writing a shader which always "consumes" n adjacent pixels for rendering.

Do you have a better idea in mind? This feels a little bit hacky.

woelper:
I would probably just convert them to rgb8 or rgba8 from the DynamicImage I think.

B-LechCode:
> I would probably just convert them to rgb8 or rgba8 from the DynamicImage I think.

Okay, maybe we should then limit the type conversion stuff to RGBA and Grey (for 8, 16 and 32 float bit depths) only? But maybe this degree of freedom is a good thing just to convert images?
Or should we make the image loader to directly convert to one of those 6 types automatically? - That's a thing we just need to specify, then I'll write the code for the texture-wrapper.

Please just think about it and tell me which option fits best :)

Stoppedpuma:
@B-LechCode I've merged my branch with some additional tests you can use for this PR. There's a shell script located in the tests folder which creates a bunch more for some possible edge cases like odd resolutions, yuvs, bit depths, formats, etc.

B-LechCode:
> Not sure if anyone is aware of this one yet, but the zoom preview seems to be able to draw above overlays.
> 
> image

Cool! Thanks for sharing that info :)

woelper:
I think this is because it draws last and does not sort with egui. How about we hide the info if file browser is open?

B-LechCode:
> I think this is because it draws last and does not sort with egui. How about we hide the info if file browser is open?

What about a second egui context which renders after? Could both be a possible solution.

woelper:
That could be an option if it works with notan. We'd need to deal with multiple contexts and UI and determine when to use what, which might not be obvious. An advantage would be that we could use it to draw elements auch as gizmos last.

Stoppedpuma:
> I think this is because it draws last and does not sort with egui. How about we hide the info if file browser is open?

The problem I think here is it removes the overlay / window feel which makes the "file manager" feel more like a "file overlay". I think this doesn't make much sense to do without making other windows separate from the main instance as they also have the same kind of feeling. I would prefer avoiding this direction as well since it's already a bit of a headache on tiling window managers dealing with other applications creating additional windows for something which could easily integrate into the main window UX and UI wise.

Stoppedpuma:
@woelper Few UX things I would like to add to your changes, copy and paste (I'm sure this is obvious and already planned), and the ability to right click a path name on the breadcrumbs bar to copy the path to the clipboard, a notification would be nice for showing it was copied as well!

B-LechCode:
> That could be an option if it works with notan. 

Tried to add a second egui context. It doesn't work - the UI gets really broken and it doesn't fix the issue.
So we need another solution on this. Any ideas, @woelper and @Stoppedpuma?

Currently L8 and RGBA8 are natively supported, if images with other types are passed, they get converted to RGBA8. 
So fixing this might be the last thing keeping us from merging?

Stoppedpuma:
> Any ideas, @woelper and @Stoppedpuma?

Is making the preview window embedded a possibility instead of having it as a separate drawable? If we can't find a way to work around this then I can probably think of some hack which looks okay to the user with basic animations, I'll hold back on this until woelper get a word in though.

Stoppedpuma:
I'll be pushing some basic file manager changes to fit closer to the reference design shortly, I did come across this though:

image

The folders seem to become offset starting at the 3rd folder.

woelper:
> Hi,
> 
> I got an error while stepping through the test images, see below. Image is: format-test.dds. I added a truncation mechanism to supress that error. But I'm curious: why is the buffer too large?
> 
> [2024-12-05T06:36:42Z DEBUG oculante::image_loader] Detected as "Microsoft DirectDraw Surface" dds [2024-12-05T06:36:42Z DEBUG oculante::image_loader] matching 'dds' [2024-12-05T06:36:42Z DEBUG oculante::utils] Got a frame receiver from opening image [2024-12-05T06:36:42Z DEBUG oculante::utils] Received image in 42.528301ms [2024-12-05T06:36:42Z DEBUG oculante] Got frame: Still ... [2024-12-05T06:36:42Z DEBUG oculante] Received image buffer: (1000, 750) [2024-12-05T06:36:42Z DEBUG oculante::texture_wrapper] Updating or creating texture with new size. [2024-12-05T06:36:42Z DEBUG oculante::texture_wrapper] Rgba8 thread 'main' panicked at src/texture_wrapper.rs:165:27: Problem generating texture: "Texture type Rgba32 with 3008000 bytes, when it should be 3000000 (width: 1000 * height: 750 * bytes: 4)"

This is odd. At this point, everything related to the original format should be gone and we should just deal with a regular `image` buffer. I suspect Rgba32 for notan is what `image` calls Rgba8 right? Otherwise I think this should work just like any other buffer.

woelper:
As for the file browser, let's just not draw the info panel when it is open.

Stoppedpuma:
Two things which come stick out to me from the recent file manager changes:

1: directories have a small chance to be missing completely if there's not enough horizontal room (possibly images too?)
2: Large directory buttons look quite bad, I can redesign this if we're going forward with large directory buttons.

B-LechCode:
> This is odd. At this point, everything related to the original format should be gone and we should just deal with a regular `image` buffer. I suspect Rgba32 for notan is what `image` calls Rgba8 right? Otherwise I think this should work just like any other buffer.

I fixed it by truncating the buffer, so I suspect the image loading to add some extra bytes.

Stoppedpuma:
Looks good to me on the file manager side, there's a few small things below which would be nice fixes / improvements prior to merge but are not blockers. Will also update #515 with these.

Issues:

- [ ] Thumbnail size math is incorrect somewhere, even though height is currently 80, it outputs 88. The size we should match is 120x90.
- [ ] The scrollbar doesn't have a background in the file manager.
- [ ] Folder icons are placeholders (This one is on me, don't wait for me if it's not ready by the time you want to merge).
- [ ] Directory path input icon is a place holder (Also on me, don't delay for me, feel free to just merge with out it).
- [ ] Bookmark icons can overlap with text, make this scrollable if it's too long.

Improvements:

- [ ] @woelper pinging you here for this because you might see this as a blocker, failed thumbnails do not retry (I think).
- [ ] Text longer than can be shown doesn't scroll
- [ ] Adjust preview to be based off the centre of the image instead of the left side.
- [ ] Copy and Pasting in the directory path area
- [ ] Right clicking a directory on the breadcrumb bar should copy path to clipboard (a notification could be nice here too).

woelper:
> Update: There seems to be one issue which **must be fixed prior to merge**, things still seem to disappear depending on the window size. A video showing this is attached below.

Sorry, I don't understand what the issue is here - the window seems to be resized and the items are wrapped to a new row. What needs to be fixed?

Stoppedpuma:
There are three images here that are purple, if resized in a specific way, only two will be present.

image
image

Another example can be seen here:

https://github.com/user-attachments/assets/fc9ccf86-7aa0-4182-a8f4-b42e2d024661

B-LechCode:
> I'm likely confused here but wasn't one of the reasons for this PR for native format support such as rgba32f? Does your fix mean allowing these formats instead of conversion?

You're right: I forgot about rgba32f - working on it. Currently no images other than 8 bits per channel get passed.
16 Bit per Channel formats are missing, too.
I will implement the support for these formats for TextureWrapper, but at some point the conversion to 8 bit on image loading needs to be removed for testing and for a real benefit of this (obviously).

> One more thing I can immediately point out, a user complained about the zoom window "wrapping" the image. This should probably change as it looks quite odd. Example with `checker.png`:

Working on it!

Stoppedpuma:
There still seems to be some artifacts on hover:

image
image

Stoppedpuma:
@B-LechCode is the old picker still present somewhere? I'm bringing this up because in the past when this PR was first opened it seemed as if the old and new picker were visually having a race condition and appearing over one another. Now that is fixed though but I found another similar issue, the pickers corners seem to round like the old picker whenever hovering over transparency. This makes me think there's still something of the old picker around somewhere or if you managed to round the new one then it doesn't work all the time.

Video showing this (Watch the corners):


https://github.com/user-attachments/assets/278d659a-cc4c-46c0-a1f0-dd32b467f759

Stoppedpuma:
@woelper Is there any way that you could make the channel notification be centred at all times? When panels are open it looks a bit odd.

image

woelper:
> @woelper Is there any way that you could make the channel notification be centred at all times (move dynamically based on if panels are opened)? When panels are open it looks a bit odd.
> 
> image
> 

I am actually using a method that promises to get the rect minus the side panels. Maybe something is off with the egui / notan integration since this seems to return the whole screen rext.

B-LechCode:
> @B-LechCode is the old picker still present somewhere? I'm bringing this up because in the past when this PR was first opened it seemed as if the old and new picker were visually having a race condition and appearing over one another. Now that is fixed though but I found another similar issue, the pickers corners seem to round like the old picker whenever hovering over transparency. This makes me think there's still something of the old picker around somewhere or if you managed to round the new one then it doesn't work all the time.
> 
> Video showing this (Watch the corners):
> round.mp4

@woelper thanks for fixing this!


> One more thing I can immediately point out, a user complained about the zoom window "wrapping" the image. This should probably change as it looks quite odd. Example with checker.png

@Stoppedpuma I'm working on this. I hope to get this to work within the next week.

src/image_editing.rs
```diff
@@ -921,215 +1116,293 @@ impl ImageOperation {
...
+                            let i = img.clone();
+                            let mut data = i.into_raw();
+                            libblur::stack_blur(
+                                data.as_mut_slice(),
+                                img.width() * 4,
+                                img.width(),
+                                img.height(),
+                                (*amt as u32).clamp(2, 254),
+                                libblur::FastBlurChannels::Channels4,
+                                libblur::ThreadingPolicy::Adaptive,
```
awxkee, line 1134: libblur's have gated feature _image_ with prebuilt handling `DynamicImage` if this makes sense in latest versions
https://github.com/awxkee/libblur/blob/03341a4e4df05d88bf61f69187cbe2c16cee7770/src/lib/stack_blur_image.rs#L46.

Just to confirm because it is not clear from context: it is a mistake that immediately leads to artifacts to perform blur/scaling/any blending operation on the image with unpremultiplied alpha. The `DynamicImage` type from `image` crate always considers an image as unpremultiplied, therefore to perform any blending operation you have to premultiply alpha first. Thus this leads to question, does here `DynamicImage` holds an image with already premultiplied alpha?

B-LechCode:
@Stoppedpuma 
I fixed (at least I hope :D) the picker:
Screenshot_20250106_000456

Can you confirm?

Stoppedpuma:
> @Stoppedpuma I fixed (at least I hope :D) the picker: Screenshot_20250106_000456
> 
> Can you confirm?

1px off it seems, applies everywhere.

image

B-LechCode:
> > @Stoppedpuma I fixed (at least I hope :D) the picker: Screenshot_20250106_000456
> > Can you confirm?
> 
> 1px off it seems, applies everywhere.
> 
> image

Thought the same way, but for the images I tested it's in the pixel center.

Stoppedpuma:
> Thought the same way, but for the images I tested it's in the pixel center.

Maybe it only applies to images with odd resolutions? This is the image I'm using for testing, you'll have to screenshot and zoom in to see this as it's extremely small.
example-1

I've also tested on just a random image hovering over a blue line, it's inaccurate there as well.

B-LechCode:
> > Thought the same way, but for the images I tested it's in the pixel center.
> 
> Maybe it only applies to images with odd resolutions? This is the image I'm using for testing, you'll have to screenshot and zoom in to see this as it's extremely small. example-1
> 
> I've also tested on just a random image hovering over a blue line, it's inaccurate there as well.

Okay, will to look into this sometimes next week. I can't see something odd with the image you provided. So I will need to generate some edge case test images.

Thanks for looking into it!

Stoppedpuma:
> > > Thought the same way, but for the images I tested it's in the pixel center.
> > 
> > 
> > Maybe it only applies to images with odd resolutions? This is the image I'm using for testing, you'll have to screenshot and zoom in to see this as it's extremely small. example-1
> > I've also tested on just a random image hovering over a blue line, it's inaccurate there as well.
> 
> Okay, will to look into this sometimes next week. I can't see something odd with the image you provided. So I will need to generate some edge case test images.
> 
> Thanks for looking into it!

I think I might have found the issue? I think the underlying problem here is actually that the zoom picker doesn't fill the entire pixel, it's centred on a 1x1 pixel but shows the outer corners of the pixel. This would make sense with why the edges of the image I posted were like this as well. Video showing this in-case I'm just being stupid (Ignore any artifacts, wf-recorder likes to create those):

[pixel.webm](https://github.com/user-attachments/assets/39c34b4c-e95a-4b2f-b803-3ef65fc1ba47)

B-LechCode:
> I think I might have found the issue? I think the underlying problem here is actually that the zoom picker doesn't fill the entire pixel, it's centred on a 1x1 pixel but shows the outer corners of the pixel. This would make sense with why the edges of the image I posted were like this as well.

What do you mean by "doesn't fill the entire pixel"?

Stoppedpuma:
> What do you mean by "doesn't fill the entire pixel"? Having the currently selected pixel under the crosshair is the way I would intend it - but maybe I missed something?

That makes sense on the surface, but in this case it doesn't necessarily because it's not visible under normal circumstances. The main problem this brings is on the edges of images and on one pixel lines where if the picker is zoomed in on it looks like there's another pixel on both up and down, or left and right if that makes sense? In my mind, the centre of the picker should fill the entire pixel, not majority of the hovered pixel minus the corners?

> Currently it's working with discrete pixel coordinates, so it's not moving subpixel wise.
That's fine, that's how this should be.

Stoppedpuma:
@B-LechCode *Theoretically* I fixed this? Please make sure this change doesn't break something somewhere.

Stoppedpuma:
> @Stoppedpuma Seems to be ok, maybe we could use a gradient brush with some alpha for the crosshair?

I agree! Give things a shot and I can fine tune some values if needed. Thanks!

One more thing I did want to ask is if you had any other changes pending? I remember you mentioning things being converted to RGB8 or something around those lines which you wanted to fix?

B-LechCode:
> I agree! Give things a shot and I can fine tune some values if needed. Thanks! 
> One more thing I did want to ask is if you had any other changes pending? I remember you mentioning things being converted to RGB8 or something around those lines which you wanted to fix?

@Stoppedpuma Yes, the 16 Bit types need to be implemented. 
RGB without Alpha is not supported by Notan, so this is currently converted to RGBA - in the case of u8 and f32 to rgba8.
Will change this to automatically fit the right depth.

woelper:
> > I agree! Give things a shot and I can fine tune some values if needed. Thanks!
> > One more thing I did want to ask is if you had any other changes pending? I remember you mentioning things being converted to RGB8 or something around those lines which you wanted to fix?
> 
> @Stoppedpuma Yes, the 16 Bit types need to be implemented. RGB without Alpha is not supported by Notan, so this is currently converted to RGBA - in the case of u8 and f32 to rgba8. Will change this to automatically fit the right depth:
> 
>     * rgb8 -> rgba8
> 
>     * rgb16 -> rgba16
> 
>     * rgbf32 -> rgbaf32
> 
> 
> @woelper In the future the automatic conversion when the images are loaded should be changed in my opinion --> Would then change nothing and maybe we can do a gray value min/max scaling via display shader then. What do you think about this? As a bonus, the displayed pixel values are the real ones of the image file then (maybe with a toggle option).
> 
> Automatic conversions in the Edit tab should then only run if the host has enough memory left - otherwise ask the user to proceed?

I think we can release this as it is now. For _display_ purposes, in the future I'd love to do the following:

 * rgb8 -> rgba8
 
 * rgb16 -> rgbaf32

* rgbf32 -> rgbaf32

And then send any image with more than 8 bpp as they are, and use a shader for display (maybe with tonemapping).

At the moment, we are still sending some images like exr, raw and hdr as rgba8. We could send anything after loading that is > 8bpp as 32bpp and we add a shader in texturewrapper.rs to make sure it is displayed nicely - you might have to help me with that @B-LechCode as you know the code best - what do you think?

Stoppedpuma:
> I think we can release this as it is now. 

I'm half and half on this, while we are about two months late to a release following the "every 1-2 months" part of the readme, this update also happens to be the size of two or possibly even three releases and should probably receive complete implementations of the important things since you wanted to have 0.9.3 be spent on rewriting a bunch of things. On the other hand, this does feel good enough to release without and catastrophically large annoyances or issues. @B-LechCode If you had to estimate, how long do you think it would take you to finish your your changes? If you can get this out within around a week or so, I say we wait for your changes, but maybe that's just me? I think this would make it easier to do any rewriting (if needed) of Simons code as it can all be done at once instead of 3/4ths now and the other quarter later?

@woelper if you are wanting to just get a release over with then there's a few things I would like to request make it into the next release as these are quick, easy fixes with large benefits.

#530
#576
#588
#597

> For display purposes, in the future I'd love to do the following:

    rgb8 -> rgba8

    rgb16 -> rgbaf32

    rgbf32 -> rgbaf32

Not sure how I feel about this one, if anything it's just unnecessary confusion on the users end? I would expect an rgb8 image to be shown as rgb8 instead of rgba8 as this would lead me to think at there's possibly an issue with how the image was encoded instead of it being Oculante's doing at first without any prior knowledge of this "change".

Stoppedpuma:
> I don't think the user will notice any of this. This is just the texture type. The user will still see their rgb8 image as rgb8 in the info window - it will just render as an rgba8 texture.

If we release now, users from a earlier version will also not notice the difference if they open an image in higher bit depth. It will look as before because it is tonemapped in software and sent as LDR. Later it would be nice to send images with higher bit depth the way they are and use a shader to compress/map them to the visible spectrum but we can always do that later.

This sounds good to me, I was thinking of this the other way around where the info part would be displayed as rgba8 but the image was still rendered as rgb8. As long as it's displayed to the user correctly both on the image side of things as well as the information side of things, this should be fine. Regarding higher bit depths, this and HDR are also something pretty high on my list! Looking forward to it!

B-LechCode:
> > > I agree! Give things a shot and I can fine tune some values if needed. Thanks!
> > > One more thing I did want to ask is if you had any other changes pending? I remember you mentioning things being converted to RGB8 or something around those lines which you wanted to fix?
> > 
> > 
> > @Stoppedpuma Yes, the 16 Bit types need to be implemented. RGB without Alpha is not supported by Notan, so this is currently converted to RGBA - in the case of u8 and f32 to rgba8. Will change this to automatically fit the right depth:
> > ```
> > * rgb8 -> rgba8
> > 
> > * rgb16 -> rgba16
> > 
> > * rgbf32 -> rgbaf32
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @woelper In the future the automatic conversion when the images are loaded should be changed in my opinion --> Would then change nothing and maybe we can do a gray value min/max scaling via display shader then. What do you think about this? As a bonus, the displayed pixel values are the real ones of the image file then (maybe with a toggle option).
> > Automatic conversions in the Edit tab should then only run if the host has enough memory left - otherwise ask the user to proceed?
> 
> I think we can release this as it is now. For _display_ purposes, in the future I'd love to do the following:
> 
>     * rgb8 -> rgba8
> 
>     * rgb16 -> rgbaf32
> 
>     * rgbf32 -> rgbaf32
> 
> 
> And then send any image with more than 8 bpp as they are, and use a shader for display (maybe with tonemapping).
> 
> At the moment, we are still sending some images like exr, raw and hdr as rgba8. We could send anything after loading that is > 8bpp as 32bpp and we add a shader in texturewrapper.rs to make sure it is displayed nicely - you might have to help me with that @B-LechCode as you know the code best - what do you think?

@woelper I'm on it, maybe I will be ready within the next week. What do you think about @awxkee's comment? I don't know the matter, but maybe you.

@Stoppedpuma I understand your desire for hdr stuff. To implement this with texturewrapper, you (maybe @woelper, too) will need to specify what to implement - that's outside my knowledge. 
Usually I use two techniques: 
1. Scaling the pixel value from min to max.
2. Moving a 8 Bit window over the Histogram and clip everything under/above.
I guess there's way more?
Nevertheless I suggest to implement this with a new issue after the next release.

My suggestion for my next activities:
- Automatic conversion of:
  rgb8 -> rgba8
  rgb16 -> rgbaf32
  rgbf32 -> rgbaf32
  Adding L16 support
- Code cleanup
- Bugfixing if something comes up

The next release will be a major leap forward! With this branch I was able to display grayscale images with 3.6 GPixels and zooming smooth through them :)

awxkee:
> * Scaling the pixel value from min to max.
> 
> * Moving a 8 Bit window over the Histogram and clip everything under/above.
>   I guess there's way more?
>   Nevertheless I suggest to implement this with a new issue after the next release.

To render HDR image you should create surface that supports required OETF curve PQ or HLG. If HDR surface is not available or diplay is not HDR, or you don't want a hardware surface then you have to apply OETF yourself and perform tone mapping to reduce HDR range into SDR.

Stoppedpuma:
> a gradient brush with some alpha for the crosshair?

@B-LechCode Did you want to get in these changes for the picker as well? Or do you want me to open an issue to keep track of it for 0.9.3 instead?

B-LechCode:
> > a gradient brush with some alpha for the crosshair?
> 
> @B-LechCode Did you want to get in these changes for the picker as well? Or do you want me to open an issue to keep track of it for 0.9.3 instead?

Let's do this in the future. 

> > * Scaling the pixel value from min to max.
> > * Moving a 8 Bit window over the Histogram and clip everything under/above.
> >   I guess there's way more?
> >   Nevertheless I suggest to implement this with a new issue after the next release.
> 
> To render HDR image you should create surface that supports required OETF curve PQ or HLG. If HDR surface is not available or diplay is not HDR, or you don't want a hardware surface then you have to apply OETF yourself and perform tone mapping to reduce HDR range into SDR.
> 
> Formally: clamp is tone mapping, perhaps the worst one.

Thank you! As an image processing guy clamping and linear mapping are used frequently :) But it's ugly from a photographs (or similar) perspective.

awxkee:
> > 
> 
> Thank you! As an image processing guy clamping and linear mapping are used frequently :) But it's ugly from a photographs (or similar) perspective.

@B-LechCode 
The thing is that there is no defined standard how to perform tone mapping.
There is only _recommendation_ from ITU-R Rec.2408 as some variant of Reinhard Extended tone mapper which also used by Chromium and Safari ( perhaps Edge also ) as fallback on devices with no HDR surface support.

Here is some example:
```c++
  const float Ld = contentMaxBrightness / whitePoint;
  const weightA = (displayMaxBrightness / whitePoint) / (Ld * Ld);
  const weightB = 1.0f / (displayMaxBrightness / whitePoint);
 const float lumaPrimaries[3] = {0.2627f, 0.6780f, 0.0593f};

  for (uint32_t x = 0; x < width; ++x) {
    float r = targetPlace[0];
    float g = targetPlace[1];
    float b = targetPlace[2];
    float inLight = lumaPrimaries[0] * static_cast<float>(r) + lumaPrimaries[1] * static_cast<float>(g)
        + lumaPrimaries[2] * static_cast<float>(b);
    if (inLight == 0) {
      continue;
    }
    float scale = (1.f + vWeightA * inLight) / (1.f + vWeightB * inLight);
    targetPlace[0] = std::min(r * scale, 1.f);
    targetPlace[1] = std::min(g * scale, 1.f);
    targetPlace[2] = std::min(b * scale, 1.f);
    targetPlace += 3;
  }
```

Where if not defined `displayMaxBrightness = 250`, `whitePoint = 203` and `contentMaxBrightness = 1000`. Those values used in Chromium, Safari uses close values but slightly different.

Content brigntess might be retrieved from avif CLLI box or from libheif.

```c++
  if (heif_image_has_content_light_level(img.get())) {
    heif_content_light_level light_level = {0};
    heif_image_get_content_light_level(img.get(), &light_level);
    if (light_level.max_content_light_level != 0) {
      contentMaxBrightness = light_level.max_content_light_level;
    }
  }
```

However, actually it is still not mandatory to do this exactly that way. Thus for example you can choose some other tone mapper from here https://64.github.io/tonemapping/ or/and perform tone mapping in Oklab, CIE LAB, CIE LUV colorspaces and it is still will be correct, and performing tone mapping in some perceptual colorspace might be often a good idea, but your results then be different from Chromium.

Anyway, if it is possible to check that your framework support HDR surfaces as VK_COLOR_SPACE_HDR10_ST2084_EXT/VK_COLOR_SPACE_HDR10_HLG_EXT it will be better to made them and do not perform tone mapping.

src/image_editing.rs
woelper: > libblur's have gated feature image with prebuilt handling DynamicImage if this makes sense in latest versions

Thanks! I will give it a try! On many libraries that expose `image`, this has been often a pain as the `image` library in this application needs to match the ones used in the library. This is often difficult to upgrade.

> Just to confirm because it is not clear from context: it is a mistake that immediately leads to artifacts to perform blur/scaling/any blending operation on the image with unpremultiplied alpha. The DynamicImage type from image crate always considers an image as unpremultiplied, therefore to perform any blending operation you have to premultiply alpha first. Thus this leads to question, does here DynamicImage holds an image with already premultiplied alpha?

This is a great observation. I hope I am understanding you correctly - I am doing no manipulation on the opened image. Actually one of the first use cases for this program was to identify images that have been incorrectly saved as premultiplied/unpremultiplied. I can not make any assumption on the images loaded, and I don't think it would be good to automatically convert or change the image data. There are "multiply by alpha" and "divide by alpha" operators though which could potentially be used before any other operation.

awxkee: @woelper 

This is a mistake then, if image have real alpha ( not a constant one ) and original color info blurring/resizing will immediately corrupt an image and this is quite tricky knowledge for many users. 

There is an example how that works:

Original image
366858929-20244122-fb60-4dcd-bd18-a82cd1aa4ed1

Blurred image

366859927-99a7f785-1cb6-4e9e-9e85-00d6210a0a9c

Effect the same for any blending operation: blurring, resizing, blend over etc.

Unfortunately, this is a bad practice to ignore alpha multiplication for blending operations. Any rendering software usually handles alpha multiplication out of the box, because there is actually no way to correctly blend colors without alpha weighting.

>  There are "multiply by alpha" and "divide by alpha" operators though which could potentially be used before any other operation.

At least if you wish to delegate this to user it might be useful to explain to user when and how to use it? Because it is needed to blending operations, and image formats always stores images as not multiplied by alpha.

Otherwise, if there is some sense for you: it is possible to do fast scanning alpha is not constant and there is some point to do alpha weighting first:

```rust
pub(crate) fn has_non_constant_cap_alpha<
    V: Copy + PartialEq + BitXor<V, Output = V> + 'static + AsPrimitive<J> + 'static,
    J: Copy + AddAssign + Default + 'static + Eq + Ord,
    const ALPHA_CHANNEL_INDEX: usize,
    const CHANNELS: usize,
>(
    store: &[V],
    width: usize,
    stride: usize,
) -> bool
where
    i32: AsPrimitive<V>,
    u32: AsPrimitive<V> + AsPrimitive<J>,
{
    assert!(ALPHA_CHANNEL_INDEX < CHANNELS);
    assert!(CHANNELS <= 4);
    if store.is_empty() {
        return false;
    }
    let first = store[ALPHA_CHANNEL_INDEX];
    let mut row_sums: J = 0u32.as_();
    for row in store.chunks_exact(stride) {
        let row = &row[..width * CHANNELS];
        for color in row.chunks_exact(CHANNELS) {
            row_sums += color[ALPHA_CHANNEL_INDEX].bitxor(first).as_();
        }
        if row_sums != 0.as_() {
            return true;
        }
    }

    let zeros = 0.as_();

    row_sums.ne(&zeros)
}
```

awxkee: > Thanks! I will give it a try! On many libraries that expose `image`, this has been often a pain as the `image` library in this application needs to match the ones used in the library. This is often difficult to upgrade

You could just copy method that binds Libblur to image crate then.

woelper: Thank you - this is great input!

What would be your recommendation then to make this more user friendly? Should we premultiply the image at the start of image processing to make sure all operations happens on a premultiplied image?

It is probably not too serious right now as we don't have any advanced things or compositing / blending operations, but you are right that we should start with good defaults and do things the right way!

awxkee: I believe to be more user friendly it will be better to premultiply when filter chain has some filter that requires it ( and unpremultiply at end of the chain or before saving ). 
Multiplication by alpha is destructible and irreversible, and resource costly. Thus for _all operations_ has its own drawbacks. 

Here is an example:
RGBA(1,2,3,2) -> (RGB(1,2,3)*2)/255 = RGBA(0,0,0,2).

Therefore, after multiplication, some original color information is lost forever. However, from an implementation standpoint it is much easier simply premultiply for any filter.

Thus, it depends on how much effort you want to invest in it and your goal for color preservation.
You could always perform multiplication, or divide the filters into three groups: required, not required, or conditionally required, and decide accordingly.

B-LechCode:
@woelper @Stoppedpuma My work with implementing the various conversions are done and tested with and without tiling (type conversion operator is really neat for this).
From my side it's done for this pr and I'm waiting to do the final testing and reviewing with you.

Please ping me if you need something :)

Stoppedpuma:
I think this should be good on my end as well, I've tried to intentionally break things and I haven't encountered anything which isn't already reported.

Small list of things which should be double checked or implemented on @woelper's end before merge:

Making sure all build flags work
Removing useless / outdated comments
Quick pass on code to look for any potential issues or things which aren't pretty

Woelpers adjustments to the view original / modified button
Implementing #530 (CJK language support), compressed to not bloat binary (**Not a requirement but nice to have**)
Decide if #562 should be merged or not

woelper:
Language support is in, but will bloat binary. I tested compression, but that makes startup less snappy (noticeable delay in startup). I made a cleanup task to use system fonts. Also added test files for Arabic and Simplified Chinese.

B-LechCode:
> Language support is in, but will bloat binary. I tested compression, but that makes startup less snappy (noticeable delay in startup). I made a cleanup task to use system fonts. Also added test files for Arabic and Simplified Chinese.

What algorithm did you use for compression? Zstd should be quite fast if usable.

Stoppedpuma:
> > Language support is in, but will bloat binary. I tested compression, but that makes startup less snappy (noticeable delay in startup). I made a cleanup task to use system fonts. Also added test files for Arabic and Simplified Chinese.
> 
> What algorithm did you use for compression? Zstd should be quite fast if usable.

I believe it was xz which is several magnitudes slower than say zstd-fast.

Stoppedpuma:
@woelper Bug I just found which should be fixed prior to merge, edits no longer seem to load.

B-LechCode:
One thing I found: the Histogram is currently based on rgba8, when should we change this?
Next version? Would create an issue for this then.

Stoppedpuma:
@B-LechCode Opened an issue at #606 for this to be in 0.9.3!

Stoppedpuma:
Two more weird behaviours I've found

1: The colour channel still be changed even when no images are present by simply pressing e, then a colour channel key.

2: More weirdness in the file manager depending on the size (Thumbnails shifting around, not the artifacts, the artifacts are an encoding issue with wf-recorder):

[scroll.webm](https://github.com/user-attachments/assets/33e73910-4409-4267-ba36-7996226d1b5b)

woelper:
> 1: The colour channel can still be changed even when no images are present by simply pressing e, then a colour channel key.

Nice find!

I think 1 is not bad actually. We could forbid some actions if no image is loaded, but maybe the user wants go into alpha mode before loading the first image or enable info or editing in advance.

I've managed to reproduce the file manager issue too. It seems to appear just at the right spot between resizing the window. I'll see if I can change that, otherwise I'd keep it and improve the file manager more in the next release, hopefully with proper resizing in all directions.

woelper:
@Stoppedpuma could you take a look if the "snapping" resize would work in the latest commit?

Stoppedpuma:
> @Stoppedpuma could you take a look if the "snapping" resize would work in the latest commit?

It works, it can get a bit weird in some places but for the most part it's fine. The two things which stick out to me is the resizing being far from the mouse (this might not be fixable and is okay), and the other which should be fixed is it getting stuck and not being able to resize it again unless if from near the close button. Video showing both of these behaviours:

[fm.webm](https://github.com/user-attachments/assets/2e2eb57e-9fc7-48f1-9800-bdc0391124dd)

woelper:
Thanks everyone! I'll merge this and we'll address remaining issues separately.

PR #496, opened 2024-10-29
Search in filebrowser

woelper:
Initial search field for searching directory contents. No metadata support for now.

image

### DIFF
--- a/src/filebrowser.rs
+++ b/src/filebrowser.rs
@@ -133,9 +133,11 @@ pub fn browse<F: FnMut(&PathBuf)>(
 
         if ui
             .add(
-                egui::Button::new(format!("{SEARCH}"))
-                    .rounding(5.)
-                    .min_size(vec2(0., 35.)), // .shortcut_text("sds")
+                egui::Button::new(
+                    RichText::new(format!("{SEARCH}")).color(ui.style().visuals.selection.bg_fill),
+                )
+                .rounding(5.)
+                .min_size(vec2(0., 35.)), // .shortcut_text("sds")
             )
             .clicked()
         {
@@ -154,9 +156,12 @@ pub fn browse<F: FnMut(&PathBuf)>(
         }
         if ui
             .add(
-                egui::Button::new(format!("{CHEVRON_UP}"))
-                    .rounding(5.)
-                    .min_size(vec2(0., 35.)), // .shortcut_text("sds")
+                egui::Button::new(
+                    RichText::new(format!("{CHEVRON_UP}"))
+                        .color(ui.style().visuals.selection.bg_fill),
+                )
+                .rounding(5.)
+                .min_size(vec2(0., 35.)), // .shortcut_text("sds")
             )
             .clicked()
         {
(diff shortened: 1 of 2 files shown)


Stoppedpuma:
Seems fine upon first use! I Do have one question though, is the previous directory button still needed? I think it feels unnecessary and just takes up room?  

There are two possible design routes which I can think of off the top of my head though which I would like to have in-place prior to merging.

1 (a more centred feeling version of yours):

image

2 (original intention, better for long searches:

image

Stoppedpuma:
Search bar has incorrect colours, match these:

Dark: #0D0D0D
Light: #E6E6E6

Stoppedpuma:
Typing indicator seems to have a light mode colour while under dark mode (assuming the blue isn't intentional).
image

woelper:
This is the stock egui textinput. I will have to check what can be done in order to theme it (or which internal colours it uses)

woelper:
> Seems fine upon first use! I do have one question though, is the previous directory button still needed? I think it feels unnecessary and just takes up room?

This button goes one level up. I personally like to navigate that way as I can repeatedly press it to move up in the directory structure.

> 1 (a more centred feeling version of yours):

It is quite hard to match that without hardcoding pixel widths. Maybe that is possible. I could also auto-extend it based on the input length.

Stoppedpuma:
> It is quite hard to match that without hardcoding pixel widths. Maybe that is possible. I could also auto-extend it based on the input length.

Maybe the second approach would work better then? Things in egui seem to resize the entire area so I can see how this could end up making the file manager very long horizontally when combining longer searches inside longer directory paths. This would also work with having the two buttons there instead of just the search, when not searching, the search icon and the previous directory button, when in search mode, a back icon, and the x icon to clear.

woelper:
I actually managed to get it working, will push in a bit. I can expand or rearrange the whole thing if the search string grows as well. The only thing that is left is to change the background of the search bar and the cursor colour .

woelper:
Ok, I think it is ready to go!

Stoppedpuma:
Thanks! Give me around 5 minutes to do some testing and I'll approve this if I can't find anything obvious or outright broken.

Stoppedpuma:
@woelper After reopening and filling the search bar several times it seems to become vertically high.
image

woelper:
> @woelper List of problems I've encountered so far:
> 
> After reopening and filling the search bar several times it seems to become vertically high.

Nice! Unfortunately I can't reproduce this. What were your exact steps?

> 
> image
> 
> The user can't copy from or paste into the search bar

True, I can reproduce this. I think I can't do much about this.


> * is not treated as all (This can wait until filtering if you would like)

yes, I think we should make the whole regex / search part a new issue.

Stoppedpuma:
> Nice! Unfortunately I can't reproduce this. What were your exact steps?

Pressing all the new buttons about 60 times each and spamming about a thousand characters into the search bar, closing and reopening a bunch of times, something around those lines.... I'll see if I can reproduce this on video and post it here.

Also the new icons in square form are offset too far to the right.

Stoppedpuma (approved):
Thanks! Nothing stands out to me as being a critical issue so I'll merge when checks are finished.

Issue #561, opened 2024-12-20
Make the alpha channel more obvious to the user


lichspace:
**Describe the bug**
this size not show
![image](https://github.com/user-attachments/assets/0e364717-b0f8-4b00-8701-2b331dbed74b)

**To Reproduce**
open png

**Expected behavior**

**Desktop (please complete the following information):**
 - OS: Windows,
 - Version  0.9.1

woelper:
Thank you! Would you be able to attach the image here so the issue is easier to reproduce?

I am also tagging @B-LechCode here as our Big Image Expert :)

Stoppedpuma:
I believe the issue here is that the user has their colour mode set to alpha. @lichspace can you see if setting to RGBA fixes this?

[rgba.webm](https://github.com/user-attachments/assets/e3ec7357-9dea-4372-bcc0-db85f38cd1b1)

woelper:
Good call, I missed that! Maybe we want non-rgba modes to have a different color so the user notices it more easily?

Stoppedpuma:
Odd suggestion and maybe not even possible but how about embedding a informative text if there's no alpha on the alpha channel? I think I was confused by this as well at some point as it's quite easy to miss press a to change to the alpha channel.

lichspace:
> I believe the issue here is that the user has their colour mode set to alpha. @lichspace can you see if setting to RGBA fixes this?
> 
>  rgba.webm

yes change to rgba is ok。 but i dont know why  show alpha mode. sorry~

woelper:
> Odd suggestion and maybe not even possible but how about embedding a informative text if there's no alpha on the alpha channel? I think I was confused by this as well at some point as it's quite easy to miss press a to change to the alpha channel. I think just changing the colour doesn't really help and may even be more confusing, maybe a gradient or repeating pattern might work better than a different colour if the informative text is off the table?

This could work in the DynamicImage branch since we can check for the image mode. We could even prevent / disable alpha mode on images without alpha channels so we don't need a message. What is more difficult are images that are rgba but have a "solid" alpha channel. You would still see white in this case.

Maybe we could have a text somewhere if the image mode is not rgba?

Stoppedpuma:
> yes change to rgba is ok。 but i dont know why show alpha mode. sorry~

You might have accidentally pressed the keybind to switch to the alpha channel. I've opened a PR which requires holding alt by default to reduce the chances of this in the future. See #562.

Stoppedpuma:
[...]

PR #562, opened 2024-12-20
Add alt to default colour channel keybinds

Stoppedpuma:
PRing into `dynamicimage` instead of pushing since this wasn't discussed with anyone. Part of solving the confusion at #561.

Changes the default keybind for colour channels to require alt to reduce the chances of accidentally changing channels, but still being able to change channels with one hand.


woelper:
I am not sure to what this would map on OSX since there is no alt, maybe it is mapped to option? I am concerned that this makes it hard for some folks to access the shortcut.

Stoppedpuma:
I would assume option would be alt as that's what I've always found the equivalent to be? We could always do shift or just not change anything at all, I went with alt originally since people are less likely to set an equivalent keybinding to those keys combined with alt.

woelper:
I merged and tested the alt color channels PR, but after testing I chose to revert it - on mac, that is the `fn` key and feels quite awkward as a default. I think with the notification and other fixes like not saving this over restarts we can keep it as it is.

PR #616, opened 2025-01-15
fix: L16 images render as black until edit panel is open

B-LechCode:


build_mac.sh
woelper, line 1: i think this file should not be touched in this PR

woelper (changes requested):

src/texture_wrapper.rs
```diff
@@ -416,10 +417,11 @@ impl TexWrap {
                 format = notan::prelude::TextureFormat::R8;
                 pipeline = Some(create_image_pipeline(gfx, Some(&FRAGMENT_GRAYSCALE)).unwrap());
             }
-            image::ColorType::L16 => {
+            //TODO: Re-Enable when 16 Bit is implemented in notan
```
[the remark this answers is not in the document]
B-LechCode: I tried Depth16, which results in panic.
I'd stick to rgba32f and change to the native type when possible.

build_mac.sh
B-LechCode: The file was touched because I merged master and dicom to my branch.

woelper (approved):

PR #624, opened 2025-01-20
609 channel picker with shader

B-LechCode:


Stoppedpuma:
Colours don't seem to be accurate:

[colour.webm](https://github.com/user-attachments/assets/fd3cafd8-3e6c-4718-ad8a-ea14f9a51117)

This is confirmed by both oculante as well as hyprpicker.

B-LechCode:
> Colours don't seem to be accurate:
> colour.webm
> 
> This is confirmed by both oculante as well as hyprpicker.

Thank you! I see what I missed here.

Stoppedpuma:
> > Colours don't seem to be accurate:
> > colour.webm
> > This is confirmed by both oculante as well as hyprpicker.
> 
> Thank you! I see what I missed here. Edit: I was too fast - I don't see the issue. I watched the RGBA values and the red channel is constant 1.0. Seems to be ok? The Red channel is between ff and fe there, your screenshot replicates these values to all channels?
> 
> Left: build from master, Right: build from my PR - no difference visible for me :) image

Apologies, this was a late night review. I think I was confused by the hex value and other colour values being wrong. I was under the assumption that this would change the colour to update as well to that channel. Thinking about this now it wouldn't really make sense since it's just greyscale.

Stoppedpuma:
The alpha channel seems to break the black border of the zoom window.

[picker.webm](https://github.com/user-attachments/assets/b3fe9a56-e822-48c1-b0c2-cf3cd6e5e678)

B-LechCode:
> The alpha channel seems to break the black border of the zoom window.
> picker.webm

Thank you for this interesing bug. That's indeed a Problem, cause the shader doesn't do a distinction between border and normal texture.
I need to think about a possible fix and will come back then :)

Stoppedpuma (approved):
Functionally, this works fine. I'll leave actual code review to @woelper.

src/ui/top_bar.rs
```diff
@@ -104,32 +104,10 @@ pub fn main_menu(ui: &mut Ui, state: &mut OculanteState, app: &mut App, gfx: &mu
...
-                            solo_channel(img, state.persistent_settings.current_channel as usize);
-                        state
-                            .current_texture
-                            .set_image(&solo_im, gfx, &state.persistent_settings);
-                    }
-                }
+        
+        if changed_channels {            
+            if state.current_image.is_some() {
+                state.current_texture.update_color_selection(gfx, &state.persistent_settings);
```
Stoppedpuma, line 110: Since clippy is being used going forwards, this should be

        if changed_channels && state.current_image.is_some() {
            state.current_texture.update_color_selection(gfx, &state.persistent_settings);
        }

src/texture_wrapper.rs
```diff
@@ -50,7 +54,78 @@ impl TextureWrapperManager {
...
+        match channel_selection {
+            ColorChannel::Red => {
+                swizzle_mat.x_axis = one_row_vec;
+                offset_vec.w = 1.0; //Alpha constant 1.0
+            }
+            ColorChannel::Green => {
+                swizzle_mat.y_axis = one_row_vec;
+                offset_vec.w = 1.0; //Alpha constant 1.0
+            }
+
```
Stoppedpuma, line 109: Whitespace

src/texture_wrapper.rs
```diff
@@ -50,7 +54,78 @@ impl TextureWrapperManager {
...
+            ColorChannel::Rgb => {
+                swizzle_mat = Mat4::from_rotation_x(0.0); //Diag
+                swizzle_mat.w_axis = Vec4::new(0.0, 0.0, 0.0, 0.0); // Kill alpha
+                offset_vec.w = 1.0; //Alpha constant 1.0
+            }
+            ColorChannel::Rgba => {
+                swizzle_mat = Mat4::from_rotation_x(0.0); //Diag
+            }
+        }
+
```
Stoppedpuma, line 127: Whitespace

Stoppedpuma (changes requested):
Few obvious things as well as a clippy recommendation (clippy is being enforced going forwards). I'll PR some more clippy fixes into your PR for you to review soon.

Stoppedpuma (approved):

B-LechCode:
@woelper is it ok to merge, or do you want to have a look, too?

woelper:
This is great work! I hope I have some time to review tonight!

src/texture_wrapper.rs
```diff
@@ -410,17 +523,18 @@ impl TexWrap {
         Option<notan::prelude::Pipeline>,
     ) {
         let mut format: notan::prelude::TextureFormat = notan::app::TextureFormat::Rgba32;
-        let mut pipeline: Option<notan::prelude::Pipeline> = None;
+        let pipeline: Option<notan::prelude::Pipeline> =
+            Some(create_image_pipeline(gfx, Some(&FRAGMENT_IMAGE_RENDER)).unwrap());
```
woelper, line 527: Will this never fail? If not, unwrap() will lead to a panic and end the program unless in a thread.

If that makes sense in this context, you can use `.ok()` on the `Result` and turn it into an `Option`, for example
```rust
let pipeline: Option<notan::prelude::Pipeline> =
    create_image_pipeline(gfx, Some(&FRAGMENT_IMAGE_RENDER)).ok();
```

woelper (changes requested):

woelper:
What do you think if we (in the future of course) always render with a display shader that has switches depending on the image or operation?

B-LechCode:
> What do you think if we (in the future of course) always render with a display shader that has switches depending on the image or operation?

Now there's always a shader in use for rendering. I only adjust a transformation matrix for the color vectors.
Don't know it it's efficient to pack everything in one big shader?

src/texture_wrapper.rs
B-LechCode: What should I do if the Option is None?
Never had a problem here, but it's a good point - should sanitize this.

src/texture_wrapper.rs
```diff
@@ -99,8 +178,63 @@ impl TexWrap {
...
+            Self::gen_texture_standard,
+            swizzle_mask,
+            offset_vec,
+        )
+    }
+
+    fn check_union_buffer_creation(buffer_result:Result<Buffer,String>)->Buffer{
+        match buffer_result {
+            Ok(buffer) => buffer,
+            Err(error) => panic!("Problem generating union buffer: {error:?}"),
```
woelper, line 197: I think this is useless: You check the result with `match`, but you panic at the `Err`. This would be equivalent to just using `unwrap()` or `expect("Problem generating union buffer")`. The app will terminate at a problem.

I would suggest to remove this function entirely.

Here are some things I wish I knew earlier about error handling:


```rust
let res: Result<std::fs::File, std::io::Error> = std::fs::File::create("foo");
// at this point, the above could have worked or not

// terminate program if this fails, for example if there is no sane way to continue
// This is hightly discouraged
let f = res.unwrap();

// if you are inside a function that itself returns a result, you can bubble the error
// upwards with `?`. The error types need to be compatible, and `anyhow` solves this.
// You need to handle the result one level higher.
fn can_fail() -> anyhow::Result<std::fs::File> {
    let res = std::fs::File::create("foo")?;
    Ok(res)
}
// It does not make much sense in a small function like this, usually
// you have a larger enclosing function that has multiple steps that can fail.


// if you just want to execute the command and don't care if it succeeds
_ = res;
// more readable would be to call this directly:
_ = std::fs::File::create("foo"); 


// In the case you don't need the result but still want to show a log output you
// can just operate on the Err part - this saves you a match:
_ = std::fs::File::open("foobar")
    .map_err(|e| error!("This failed: {e}"));


// If you don't care about the error, but still want to deal with potential absence later,
// similar to your case:
let mut maybe_file: Option<std::fs::File> = res.ok();

// you can combine this with the above of course:

let mut maybe_file: Option<std::fs::File> = res
    .map_err(|e| error!("This failed: {e}"))
    .ok();
```

woelper (changes requested):

src/texture_wrapper.rs
```diff
@@ -99,8 +178,63 @@ impl TexWrap {
...
+
+        (uniform_swizzle_mask, uniform_offset_vector)
+    }
+
+    fn update_uniform_buffer(&self, gfx: &mut Graphics, swizzle_mat: Mat4, offset_vec: Vec4) {
+        gfx.set_buffer_data(&self.uniform_swizzle_mask, &swizzle_mat);
+        gfx.set_buffer_data(&self.uniform_offset_vec, &offset_vec);
+    }
+
+    fn gen_texture_background(gfx: &mut Graphics) -> Texture {
```
woelper, line 227: this could be 
```rust
  fn gen_texture_background(gfx: &mut Graphics) -> Result<Texture> {
      let boundary_pixels_bytes: [u8; 4] = [0, 0, 0, 255];
      Ok(gfx
          .create_texture()
          .from_bytes(&boundary_pixels_bytes, 1, 1)
          .with_format(notan::app::TextureFormat::Rgba32)
          .build().map_err(|e|anyhow!("{e}"))?)
  }
```
for example.

B-LechCode: I did the unwrap/matching on purpose - all I need in the struct ist the texture to render later and not a texture option or a result I would have to handle more downstream.
It all boils down to one question: What to do if texture creation fails?
In my thoughts, panicking was ok at this point, what do you think?

I could have used the other function I created for "normal" textures, that's right. It was late last night :D

woelper: No worries, I am just nitpicking!

If the error is recoverable at all, you should not panic, ever. Otherwise for the user, the app will just close and they have no idea what happened. So if it is possible at that point in time to display a warning, that should be preferable.

src/texture_wrapper.rs
B-LechCode: At least we would see the error in the console, that's what I thought.
Will look deeper when I got the time.

src/texture_wrapper.rs
B-LechCode: Okay, I will have a look at it, when I have spare time.

src/texture_wrapper.rs
```diff
@@ -99,8 +183,73 @@ impl TexWrap {
...
+        gfx: &mut Graphics,
+        swizzle_mask: Mat4,
+        offset_vec: Vec4,
+    ) -> Result<(Buffer, Buffer), String> {
+        let uniform_swizzle_mask = gfx
+            .create_uniform_buffer(1, "SwizzleMask")
+            .with_data(&swizzle_mask)
+            .build();
+
+        match uniform_swizzle_mask {
```
woelper, line 209: Instead of the match block, you should be able to just use `?` after line 207, for exampe
```
let uniform_swizzle_mask = gfx
    .create_uniform_buffer(1, "SwizzleMask")
    .with_data(&swizzle_mask)
    .build()?;
```

If the error types don't match, you can use `.map_err()` to fix that.

B-LechCode:
@woelper I now completely reworked the error handling. I use the Result type whenever there is a notan call in a function which has an Result return type. If approved I would now update the calls to match the Result and displaying a warning if theres an Error.

What do you think? Thank you for "nitpicking", I learned many things about error handling with that. In my opinion I should have changed this way earlier.

src/texture_wrapper.rs
```diff
@@ -31,16 +35,22 @@ impl TextureWrapperManager {
...
         //First: try to update an existing texture
         if let Some(tex) = &mut self.current_texture {
             if tex.width() as u32 == img.width()
                 && tex.height() as u32 == img.height()
                 && img.color() == tex.image_format
             {
                 debug!("Re-using texture as it is the same size and type.");
-                tex.update_textures(gfx, img);
-                return;
+                match tex.update_textures(gfx, img) {
```
woelper, line 46: Not required, just a hint:

you can just use the Err if you ignore the Ok() part in match:

```rust
if let Err(error) = tex.update_textures(gfx, img) {
    self.clear();
    error!("{error}");
    return Err(error);
}
```

woelper (approved):

woelper:
> @woelper I now completely reworked the error handling. I use the Result type whenever there is a notan call in a function which has an Result return type. If approved I would now update the calls to match the Result and displaying a warning if theres an Error.
> 
> What do you think? Thank you for "nitpicking", I learned many things about error handling with that. In my opinion I should have changed this way earlier.

Super nice, thank you for your work!!

PR #625, opened 2025-01-21
Extra language stuff

Stoppedpuma:


src/ui/theme.rs
```diff
@@ -140,7 +140,7 @@ pub fn load_system_fonts(mut fonts: FontDefinitions) -> FontDefinitions{
     debug!("Attempting to load sys fonts");
     let mut fontdb = HashMap::new();
 
-    fontdb.insert("chinese_simplified", vec![
+    fontdb.insert("asian_fonts", vec![
```
woelper, line 143: This is not strictly true. If the first match supports all asian fonts, yes. But if you ad one font for Korean and one for Japanese, only the Korean is loaded and you end up with no Japanese support. In that case, you need another set, like in the Arabic case.

src/ui/theme.rs
```diff
@@ -152,9 +152,14 @@ pub fn load_system_fonts(mut fonts: FontDefinitions) -> FontDefinitions{
         "MS Gothic",
         "Noto Sans SC",
         "PingFang SC",
+        "Source Han Sans CN",
```
woelper, line 155: This logic will load the first match. If you list different fonts here, they are mutually exclusive.

woelper (changes requested):

Stoppedpuma:
@woelper I've split and reorganised these as there were a few overlapping ones from the original implementation.

woelper (approved):

PR #629, opened 2025-01-25
Update README.md

ErebusBat:
On a brand new MacBook Sequoia 15.1 I found this was not installed and needed.

Technically XCode command line tools are also required; however this is also prerequisite for homebrew so I did not add that.


woelper (approved):
Thanks!

PR #630, opened 2025-01-29
fix: Thumbails have correct aspect ratio and reload if loaded while b…

woelper:
…eing written

### DIFF
--- a/src/ui/top_bar.rs
+++ b/src/ui/top_bar.rs
@@ -104,9 +104,11 @@ pub fn main_menu(ui: &mut Ui, state: &mut OculanteState, app: &mut App, gfx: &mu
                     });
             });
         }
-        
+
         if changed_channels && state.current_image.is_some() {
-            state.current_texture.update_color_selection(gfx, &state.persistent_settings);
+            state
+                .current_texture
+                .update_color_selection(gfx, &state.persistent_settings);
         }
 
         let label_rect = ui.ctx().available_rect().shrink(50.);
--- a/src/ui/top_bar.rs
+++ b/src/ui/top_bar.rs
@@ -311,12 +311,12 @@ pub fn draw_hamburger_menu(ui: &mut Ui, state: &mut OculanteState, app: &mut App
         ui.style_mut().override_text_style = Some(egui::TextStyle::Heading);
 
         ui.menu_button(RichText::new(LIST).size(ICON_SIZE), |ui| {
-            if ui.styled_button(format!("{MOVE} Reset view")).clicked() {
+            if ui.styled_button(format!("{MOVE}  Reset view")).clicked() {
                 state.reset_image = true;
                 ui.close_menu();
             }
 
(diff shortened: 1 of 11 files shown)


Stoppedpuma:
The "Open thumb dir" button seems to freeze Oculante until the opened file manager window is closed.

src/ui/top_bar.rs
```diff
@@ -311,12 +311,12 @@ pub fn draw_hamburger_menu(ui: &mut Ui, state: &mut OculanteState, app: &mut App
         ui.style_mut().override_text_style = Some(egui::TextStyle::Heading);
 
         ui.menu_button(RichText::new(LIST).size(ICON_SIZE), |ui| {
-            if ui.styled_button(format!("{MOVE} Reset view")).clicked() {
+            if ui.styled_button(format!("{MOVE}  Reset view")).clicked() {
```
woelper, line 314: I think it is worth checking if you can fix the spacing directly in the `styled_button` function.

Stoppedpuma: I'll look into it! Nothing really stood out to me as hard as those did, but I'll play around with it and see if I can find a good value.

Stoppedpuma: Thanks! I didn't notice the other ones as the hamburger menu ones were especially bad. Everything seems to be centred now.

Stoppedpuma:
@woelper Another thing which you should probably check out before merging is #633.

Stoppedpuma:
#633 Still happens around the actual image part, I think it's possibly just misaligned:

image

The rounded corners on thumbnails also seem to have regressed on this PR.

woelper:
I am using the exact same rect, maybe this is because the image is snapped / aligned to whole pixels?

woelper:
It seems rounding is not an option due to how the image is drawn. I can look at it, but it will likely not be possible.

woelper:
> #633 Still happens around the actual image part:
> 
> image
> 
> The rounded corners on thumbnails also seem to have regressed on this PR. This seems to be caused by [61cbce7](https://github.com/woelper/oculante/pull/630/commits/61cbce76801674212f114ad12d1bae6f539bc7e8). The correct rounding value should be 6.

Fortunately I was wrong. To me it looks good now, can you double check please?

Stoppedpuma:
This looks good to me as well! Thanks for taking the time to look into the rounding as well as the issue at #633!

One thing which sticks out to me that isn't really too much of an issue is that the new hover brightness value makes basically every images look either too bright or kind of washed out, Was there a reason for this change outside of the thumbnails?

woelper:
> This looks good to me as well! Thanks for taking the time to look into the rounding as well as the issue at #633!
> 
> One thing which sticks out to me that isn't really too much of an issue is that the new hover brightness value makes basically every images look either too bright or kind of washed out, Was there a reason for this change outside of the thumbnails?

There wasn't a reason besides me thinking that the effect was too subtle. I reverted the change to 5.

PR #637, opened 2025-02-02
Update toml cleanup

woelper:


Stoppedpuma:
I've done some small rewording to the lossless jpeg transforms section. In the future, how would you feel about just not having this section at all and just applying these lossless jpeg specific versions of the filters if the user selects them like any other format? For example, instead of having the entire lossless jpeg specific dropdown to rotate losslessly, it's instead just works like any other format by selecting the rotate filter in the filters dropdown and behind the scenes switching to use the lossless jpeg transform crop?

woelper:
> I've done some small rewording to the lossless jpeg transforms section. In the future, how would you feel about just not having this section at all and just applying these lossless jpeg specific versions of the filters if the user selects them like any other format? For example, instead of having the entire lossless jpeg specific dropdown to rotate losslessly, it's instead just works like any other format by selecting the rotate filter in the filters dropdown and behind the scenes switching to use the lossless jpeg transform crop?

I think this would be hard - this depends on the library use flag, so mixing it with another UI is complex. Also operators are non-destructive, which lossless operations are not. They are immediately applied. If you combine them in the edit stack, this would not work any more.

PR #646, opened 2025-02-16
606 extend histogram

B-LechCode:


src/histogram.rs
```diff
@@ -0,0 +1,344 @@
...
+                trans &= c1==0;
+            }
+            if USEFUL_CN > 2 {
+                let c2: usize = reducer.reduce(chunk[2]).as_();
+                bin2[c2] += 1;
+                trans &= c2==0;
+            }
+            if has_alpha{
+                let alpha: f32 = chunk[USEFUL_CN].into();
+                transparent_pixels += (alpha<=f32::EPSILON && trans) as u64;
```
awxkee, line 130: I don't think this is correct. I think there is specialization needed 
`trait AlphaTest{ 
fn is_transparent_alpha(self)
}

impl AlphaTest for u8 {
fn is_transparent_alpha(self) {
 self != 255
}
}
```
something like that.

Also it is not seems to be correct for edge cases image 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e96e6185633076b554739d50d5a210cf

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8a14359060a12b05b7485345da5239f3

Comment (awxkee) [reply src/histogram.rs:130]: I believe there is should be
```rust
    if f.is_nan() {
        f32::MAX // Or any local positive extremum, as 1, 255, 65535 etc
    } else if f32::INFINITY == f {
        f32::MAX // Or any local positive extremum, as 1, 255, 65535 etc
    } else if f32::NEG_INFINITY == f {
        f32::MIN  // Or any local negative extremum, 0 if u8, u16 type required
    } else if f.is_subnormal() {
        0. // Non-transparent value, since 1/65535 is subnormal, but for 16 bit-depth it is not transparent :)
    } else {
        f
    }
```

PR #650, opened 2025-02-26
Cache thumbnails under $XDG_CACHE_HOME

nbraud:
Previously used `$XDG_DATA_HOME`, which is per [spec] for “important” user-specific data.

[spec]: https://specifications.freedesktop.org/basedir-spec/latest/

### DIFF
--- a/src/thumbnails.rs
+++ b/src/thumbnails.rs
@@ -76,7 +76,7 @@ pub fn path_to_id<P: AsRef<Path>>(path: P) -> PathBuf {
 }
 
 pub fn get_disk_cache_path() -> Result<PathBuf> {
-    Ok(dirs::data_local_dir()
+    Ok(dirs::cache_dir()
         .ok_or(anyhow!("Can't get local dir"))?
         .join("oculante")
         .join("thumbnails"))


nbraud:
PS: I just noticed that
- e239278c07a65defb22e6e1bba496bc4111425d5 moved the configuration to user-specific data, as part of #428 
- de581a9fb41b93e73801f15a5271de9dbc508920 put the history / “recent directories” in non-essential (cached) data.

The appropriate thing to do, as far as the XDG specification is concerned, would be to move:
- the thumbnails cache under `XDG_CACHE_HOME`, (done in this PR)
- the configuration back under `XDG_CONFIG_HOME`,
- the history file under `XDG_STATE_HOME`.

I'm not writing changes for the last two right now, as (unlike the cache) care should be taken to move them over... and I'm having a severe migraine  :sweat:

nbraud:
> The XDG / FreeDesktop "Spec" here isn't really being followed here as well since this isn't really a "Linux app" (Meaning like having "native" things like GTK / adwaita / whatever else) as much as it is just an app that's "universal" (support for all major operating systems without necessarily meeting native standards) if that makes sense?

That doesn't make sense for me at all, given that all major operating systems do follow a similar distinction, and you already use the `dirs` crate whose `cache_dir` and `config_dir`[^state] functions will do the correct thing on both Linux, Windows, and (as far as I know) macOS.

[^state]: For some mildly-odd reason, `dirs::state_dir` isn't defined on Windows, even though there's a long-established convention of using the local profile for that; `state_dir().ok_or(data_local_dir())` should work though.

This isn't merely a matter of spec-compliance either, but will also cause confusing behaviour in some networked environments: for instance, using `dirs::data_local_dir` for configuration, will cause Windows users not to have their settings available when migrating to a different machine (or even just using a different one, in a typical Active Directory environment)

If needed, I can perfectly well submit PRs which do the right thing on all supported platforms... which already was the case of this one.

> I'm assuming the origin of opening this PR is because of a packaging issue

I noticed the issue while investigating a build failure, but I wouldn't have reported it if it only mattered for packaging.

nbraud:
> Oculante follows an "all in one" design where it has a portable data folder

FYI, the history / “recent folders” file already is in a different folder (as mentioned in my first comment) and IDK whether that was intentional.

nbraud:
> you have to take into consideration that Oculante is still actively designed to be portable. 

I am taking this into consideration, and my replied had been based on my own experience writing portable software.
Moreover, the `dirs` crate is already designed to deal with platform-specific data locations, i.e. it is the necessary portability layer.

> I can't really think of any applications which are supposedly portable yet make directories and or files scattered in various parts of the file system instead of just keeping everything in an all in one data folder.

This isn't an uncommon pattern for portable applications, so IDK which specific applications you are thinking of.

Stoppedpuma:
> See [filebrowser::load_recent_dir](https://github.com/woelper/oculante/commit/de581a9fb41b93e73801f15a5271de9dbc508920#diff-06db795ed1d5b5b84c91ddcd19c4cdc92b9556cf5a608d48e1577f47631b6dd0R13-R19). Note that I had already linked to the relevant commit.

Thanks! I'll fix this soon.

> Using state/cache/config dirs isn't an uncommon pattern for portable applications, so IDK which specific ones you are thinking of.

I've never seen this behaviour then. I'll be it I don't really use a lot of programs, but typically I notice that it's usually in a folder usually either in the same spot as the executable, or in a specific location like what's done with Oculante.

> I am taking this into consideration, and my replied had been based on my own experience writing portable software.
Moreover, the dirs crate is already designed to deal with platform-specific data locations, i.e. it is the necessary portability layer.

Whilst I can't and won't speak on behalf of woelper, I'm not sure if this will change. Oculante initially used an all in one folder structure I believe. Eventually, it was switched to the XDG standard after a user mentioned it. This made more sense back then since it was only really a config file, Now this has expanded. During our redesign discussions, I suggested storing LUTs and Annotations somewhere so the user didn't have to look for them each time. I proposed using a single folder which woelper agreed that he preferred that idea anyways.

At the end of this, woelper is the one with say here as it's his project. I'm personally for keeping everything all in one folder as it's not only convenient, but makes sense (at least to me) for this type of project.

nbraud:
> Whilst I can't and won't speak on behalf of woelper, I'm not sure if this will change. [...] At the end of this, woelper is the one with say here as it's his project.

Sure, I definitely can't *force* changes into oculante, nor was that my intent.

> I'm personally for keeping everything all in one folder as it [...] makes sense (at least to me)

Then please say so outright: saying it's a “personal preference” would at least avoid wasting prospective contributors' time, and make it clear this isn't based on technical grounds or actual portability issues.

nbraud:
BTW, you didn't tag @woelper so IDK whether they'd see this conversation, since you closed the PR.

Stoppedpuma:
> Sure, I definitely can't force changes into oculante, nor was that my intent.

No worries, I didn't interpret it that way, same applies the other way around in-case you interpreted me wrong, changes can be made, it's just a matter of agreement that something is an improvement.

> Then please say so outright: saying it's a “personal preference” would at least avoid wasting prospective contributors' time, and make it clear this isn't based on technical grounds or actual portability issues.

This decision was agreed upon by both of us is what I was trying to say, I was just stating my position on this up front. As mentioned above, if needed or desired by more people then changes can be made, until then though I'm not sure if changes will be made as this seems to fit our goals for now. As of this second, it may not make much sense to have a data folder which holds only a config file and thumbnail cache, but in the future this folder is planned to hold things like LUTs, Brushes, and anything else which may come up in the future.

woelper:
Hi, thanks for tagging me!

I find this to be an interesting conversation, and there are some good points being made. I think we should reopen this and at least consider moving the thumbnails to a cache directory. I think there are also good arguments for complying with standards such as in migrating user data through automated methods as mentioned above. I did not have these in mind when the code was written.

Stoppedpuma:
@nbraud @woelper How about a middle point like an "export" option? This would mean making use of standard directories by default and allow exporting things with a button in preferences. One thing I'm not stoked about is this adding complexity to something which should be pretty straight forwards, but in this case it may make sense. Regardless, here's how this would work in my mind:

This would take advantage of the automatic migration mentioned by woelper above by throwing everything into an archive file upon export, and on the opposite side, putting things into their relevant places upon import. One thing to take some careful consideration of is how we would handle importing on top of existing data. Perhaps a dialog like "Merge" or "Replace"? Togglable options may even be useful here depending on how in-depth we want this.

This would make sense for both sides of the userbase, those who expect portability, and those who just have it as a "deployed" installation and don't really care for the portability aspects.

Stoppedpuma:
@nbraud following what was mentioned above, I'm completely willing to move those mentioned above into XDG_CACHE. Since I can't re-open this PR and I already have one open that spawned from this PR, I'll be moving .efd_history (doing xdg_cache for now instead of xdg_state because I think we can drop this file in favour of the line in the volatile config but I would like to check with @woelper first) and thumbnails to ~/.cache/oculante over at #651. I've held back on pushing this for now so I can ask if you would like the contribution credit or not. You can PR into my branch so you still get a contribution or I can just co-author you in a commit. Otherwise if you don't care for the contribution, I already have the commits ready (moving these to ~/.cache/oculante/) and can just push these.

Stoppedpuma:
Okay so I'm just going to go with the changes. .last_open_directory (previously .efd_history) and thumbnails have been moved to cache as these aren't essential data can can vary between systems. Some deps have been bumped, and I'll merge this into the toml PR. I'll also update the README's uninstallation area which points towards the data and cache dirs for those who want to make sure they know where everything is.

nbraud:
Hi!  Sorry for the lack of reply, life (and lack thereof...) got in the way the last couple of weeks.

I'll try to go, check the current implementation, and send a new PR if I still find issues.

PR #651, opened 2025-02-26
Recent dir file location

Stoppedpuma:
@woelper Please review these changes just in-case I've overlooked something. I've also renamed this file from efd_history to last_open_directory as I think that makes more sense (I intentionally did open instead of opened since you have to open a file here for it to actually update I think).

Merges into #637

### DIFF
--- a/src/filebrowser.rs
+++ b/src/filebrowser.rs
@@ -17,9 +17,10 @@ use strum::IntoEnumIterator;
 
 fn load_recent_dir() -> Result<PathBuf> {
     Ok(PathBuf::from(read_to_string(
-        dirs::cache_dir()
-            .context("Can't get temp dir")?
-            .join(".efd_history"),
+        dirs::data_local_dir()
+            .context("Can't get data dir")?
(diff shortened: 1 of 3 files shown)


src/filebrowser.rs
```diff
@@ -17,9 +17,10 @@ use strum::IntoEnumIterator;
 
 fn load_recent_dir() -> Result<PathBuf> {
     Ok(PathBuf::from(read_to_string(
-        dirs::cache_dir()
-            .context("Can't get temp dir")?
-            .join(".efd_history"),
+        dirs::data_local_dir()
```
Stoppedpuma, line 20: Change sparked from here: https://github.com/woelper/oculante/pull/650#issuecomment-2686197956. If this was intentional, this can be reverted.

Stoppedpuma:
An alternative is ditching this file all together and just reading the "last_open_directory" line from `config_volatile.json` which now that I'm thinking about it would make more sense anyways since this is kind of duplicate behaviour?

PR #661, opened 2025-03-23
fix: Correctly display breadcrumbs with single chars. Fixes #653

### LINKED ISSUE #653: Breadcrumb bar doesn't handle directories named with only one character correctly
Image

woelper:


Stoppedpuma (approved):
Works fine in my testing.

PR #662, opened 2025-03-25
Update workflows

bircni:
set checkout to v4 to be faster


Stoppedpuma:
Should be fine to merge, I'm going to wait for tests to finish and make sure the binaries aren't broken in one way or another and I'll merge if all goes well.

Stoppedpuma (approved):
Seems fine to me.

Issue #643, opened 2025-02-14
Specifying images in different directores doesn't work correctly after going forwards then back


Stoppedpuma:
Steps to reproduce:

1: `oculante image.png directory/image2.png`
2:  press next, then back. The image will be of another image in the directory (assuming there's other pictures)

PR #670, opened 2025-04-08
fix(643): Launching with images in different dirs

joshuamegnauth54:
Closes: #643

Oculante replaces the current paths in its path scrubber when the directory changes. If the app is launched with images in two different directories, such as `oculante image.png a/b.png`, the aforementioned behavior clobbers the paths passed in.

---

I also removed a few clones. :smile_cat:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -203,11 +203,8 @@ fn init(_app: &mut App, gfx: &mut Graphics, plugins: &mut Plugins) -> OculanteSt
     debug!("matches {:?}", matches);
 
     let paths_to_open = matches
-        .get_many("INPUT")
+        .get_many::<String>("INPUT")
         .unwrap_or_default()
-        .cloned()
-        .collect::<Vec<String>>()
-        .into_iter()
         .map(PathBuf::from)
         .collect::<Vec<_>>();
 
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -203,35 +203,31 @@ fn init(_app: &mut App, gfx: &mut Graphics, plugins: &mut Plugins) -> OculanteSt
...
-        .into_iter()
         .map(PathBuf::from)
         .collect::<Vec<_>>();
 
     debug!("Image is: {:?}", paths_to_open);
 
     if paths_to_open.len() == 1 {
         let location = paths_to_open
-            .first()
+            .into_iter()
```
woelper, line 215: Why was this changed? I think `first()` is a bit more self explanatory.

woelper (changes requested):
Thanks for this! I just have a small nitpick, seems great otherwise!

src/main.rs
joshuamegnauth54: Hi! Yeah, I think `first()` is clearer as well, but it would require cloning the `PathBuf` since it [returns a reference](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.first).

I can revert that though since it's not a big deal or anything.

Stoppedpuma: @woelper pinging you just incase this went past you.

Stoppedpuma:
@joshuamegnauth54 just a heads up that there's a conflict here now since I've merged your other PR.

joshuamegnauth54:
Okay I'll fix it now. :smile_cat:

PR #674, opened 2025-04-11
Improve net read performance

TheLQ:
I really like this feature

Am sending large images (100 MB 6000x6000 uncompressed PNG) to the server. Takes over 3 minutes because each 100k read chunk calls `image::load_from_memory`  which understandably adds up.

This can simplify to `stream.read_to_end()`. Then only call `image::load_from_memory` once or fail. Loads nearly instantly now. 

# Alternative

This didn't work but may be interesting. 

Tried to directly pass on the stream once we get a format. But ImageReader wants Seek, and I don't know of a "Read entire stream progressively to buffer, with Seek to bounce around" ready made utility. 

Why? Support `cat img1.png img2.png | nc` which maybe was this code's future intent. 

```rust
if let Ok(format) =  guess_format(&imgbuf) {
   let new_stream = (&mut imgbuf).chain(stream);
   let new_stream = BufReader::new(new_stream);

   let mut reader = ImageReader::new(new_stream);
   reader.set_format(format);
   let frame = reader.decode()?;
```


woelper (approved):

woelper:
Thank you! ❤️

PR #676, opened 2025-04-26
fix(586): Compare list & scrubber mangles position

joshuamegnauth54:
Using both the scrubber and the compare list to go to the next or previous image may clobber the current image position. I refactored the compare list to store its position as well as removed all clones for the list.

The result is that the scrubber and compare list maintain their own positions. I assume that both lists should have their own set of images (e.g. the scrubber may be open to ~/Pictures while the compare list has whatever images the user added). This patch does not fix that issue, but I can work on that in a separate PR if needed.

### DIFF
--- a/src/appstate.rs
+++ b/src/appstate.rs
@@ -1,4 +1,5 @@
 use crate::{
+    comparelist::CompareList,
     image_editing::EditState,
     scrubber::Scrubber,
     settings::{PersistentSettings, VolatileSettings},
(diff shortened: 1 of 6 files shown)


Stoppedpuma:
Thanks for the PR! I think this works great! I apologise for the long delay that this PR has received before being reviewed. I've tested this and I think it all works great. I'll wait for woelpers review as well before merging though.

Stoppedpuma (approved):

joshuamegnauth54:
Hi! It's no problem at all. Thank you for reviewing. :grin:

PR #684, opened 2025-06-04
Allow passing options to decoders (HEIF fix)

### LINKED ISSUE #683: Large tiled heif image throws lioc box error: MemoryAllocationError(SecurityLimitExceeded) Memory allocation error: Security limit exceeded
Found while attempting to test #682 

Downloaded from [here](https://cloud.dirk-farin.de/s/9FtHJy9SmsNRWsn).

joshuamegnauth54:
Closes: #683

Decoders, like the HEIF decoder in libheif, often support tuning features such as memory usage.

For security or performance reasons, some features may be disabled or limited. Some images legitimately require more memory to decode due to their resolution.

I implemented a feature to pass options to decoders. Internally, any decoder may be supported by adding more configs. I ensured that it's extensible, but I did not implement anything beyond what is needed to fix #683.

---
~~This is a draft till I add the option to the GUI itself.~~

A good way to test this is to set a low limit then open a HEIF file. `libheif` will return an error on decoding which in turn shows that the limits are correctly applied.

You may also try it with large HEIF files, but that may cause OOM conditions if the file is huge.


Stoppedpuma:
Thanks for the PR! I'll have time to test this sometime this week hopefully and it *seems* fine to merge from just a glance at your changes. 

Would you be able to add a test file that's reasonable in file size (or adjust the create-extra-tests script accordingly to create a test file) with an applicable name? If not I'll do it myself when I have time to actually review this PR properly.

joshuamegnauth54:
Do you mean I should write a unit test for the code with a HEIF file? I can work on that soon. :smile_cat:

joshuamegnauth54:
I added unit tests for the new security limits options that reuses a HEIC file already in the repository.

Stoppedpuma:
Awesome! I think that anything that you think would affect if an image can or cannot be loaded would be worth exposing. It's a balance of exposing the right amount of advanced options but not the useless ones if you know what I mean. I would say just adapt it to what you believe should cover any cases foreseeable and we can always adjust it later if a user finds an edge case.

joshuamegnauth54:
I'll try to find documentation on the options then add them to the UI. I also fixed the bug where the setting isn't reloaded onto the UI in the commit above.

Stoppedpuma:
@joshuamegnauth54 your changes seem to have broke the input box all together just in-case you weren't aware.

https://github.com/user-attachments/assets/114e8c7b-db56-451f-8722-8236ce35e9ab

Stoppedpuma:
I've repushed my commit since your force push overwrote them. I think this should be ready after you expose any other options you believe should be present. If you're having a hard time deciding, we can just expose them all as it would probably make sense anyways for a user to have complete control just incase of edge cases.

joshuamegnauth54:
Oops. :sweat_smile: I didn't check before I force pushed. I'm going to add `Items` as an option since it will fix the image you linked above.

It might make sense to add all of the security options, as you mentioned, though I'm not really sure what they do.

Stoppedpuma:
Maybe just listing the ones you can find docs on to explain what they do. If it helps, I just threw it into gemini for some quick summarisations which may or may not be correct.

image

joshuamegnauth54:
I added all of them. :joy_cat: After doing a bit of research, I found that some programs, like ImageMagick, had issues where users had images that needed tuning different security options. It seems safest to just add all of them at once.

Stoppedpuma:
Okay so I've tested this and everything I've tested works besides that image from above still throwing the same issue, is there some flag or something we need for these values to go over their safety limits that we may be missing?

joshuamegnauth54:
I don't think that option, `iloc box`, is exposed anywhere. :thinking:

It's possible that `libheif` just doesn't expose that via its security context yet. Some developers who use `libheif` asked the upstream developers to add options based on real world images failing to decode. This may be an instance of that as well.

joshuamegnauth54:
Should that be documented anywhere in Oculante? Or maybe we could add a checkbox that says something like "Disable all security options" that sets the environmental variable for us?

Stoppedpuma:
What would you think about replacing the current options we have exposed with just a toggle button to set the variable? That would simplyify it to an option called something like "Disable all heif security limits" or something.

joshuamegnauth54:
I'm not sure what would be best. :laughing: Granularity is nice in case a user doesn't want to opt out of everything. ImageMagick exposes all of them. :thinking: 

I'll do whatever you think is best and have it done in a day or two!

Stoppedpuma:
@woelper I would like to hear if you have a preference. Currently, this PR adds several input boxes to preferences for the user to specify limits which override libheifs defaults. The problem is we have exposed all that we can find and it still leaves a few security limits unexposed. On the other hand, there's a variable that disables all security limits and we can turn the several input boxes into just a single toggle button.

I see three ways of doing this:

1. Just sricking with what we currently have in the input boxes, which does NOT cover all cases (I would like to avoid this if possible)
2. Using the single toggle button
3. Both?

I'm heavily in favour of either 2 or 3 as I want to have as many edge cases as possible. 3 is not the prettiest but I suppose allows finer grained control whilst having the disable all limits functionality? Not too sure on what I would prefer. If nothing, we technically have a 4th option of just commenting out what we currently have and using the single toggle button until if someone wants finer controls over them.

joshuamegnauth54:
I prefer options 2 or 3 too. I consider the decoder options to be advanced options (like VLC or mpc's options) so I think it's fine if it's granular.

I'll try implementing the toggle button soon. There is a chance it may not work depending on how `libheif` gets the env var. Also I may be busy this weekend so I might have to work on it by Sunday or Monday. :thinking:

Stoppedpuma:
No worries, take your time. Do you think it would be possible to hide the input boxes if the checkbox is enabled?

joshuamegnauth54:
I figured out how to disable the text inputs if the user disables all of the security options. :partying_face: I didn't test it yet so I likely have to work more on it on Monday.

Stoppedpuma:
Awesome! Thanks so much! Whilst it doesn't currently work, it does look pretty much how I imagined visually! The light mode is the only area I'd call problematic since the input boxes become practically invisible if this is enabled. I wouldn't worry about this since you've already done all the hard stuff! There's also some more UX I would like add in the future as well so feel free to leave it as is if you would like since it will likely be revisited. I've also pushed some small changes just for consistency and layout. Thanks again!

joshuamegnauth54:
It works now. :grin: Oculante must be restarted for the override to apply. The override requires libheif v1.19.8 or later. Debian seems to package that version so other distros likely have at least 1.19.8, which is a few months old.

Stoppedpuma (approved):

PR #695, opened 2025-07-04
Fix hamburger menu still showing in zen mode if borderless mode is enabled

### LINKED ISSUE #690: Add a permanent option and a temporary command to hide the hamburger menu in Zen
**Is your feature request related to a problem? Please describe.**
I'd like to hide all and any UI "distraction" from the image. Zen mode does some of that, but then the top-right hamburger menu is still visible and so is the bottom file scroll bar

**Describe the solution you'd like**
- A config option to permanently hide the hamburger menu in zen mode
- A shortcut command to temporarily hide/show this menu in zen mode
- Maybe add an alternative way to invoke this menu regardless of whether it's shown/hidden. For example, right mouse click?

**Describe alternatives you've considered**
Not sure

**Additional context**

Stoppedpuma:
Closes #690

### DIFF
--- a/src/icons.rs
+++ b/src/icons.rs
@@ -9,7 +9,6 @@ pub const CHEVRON_DOWN: &str = "\u{E91C}";
 pub const CHEVRON_UP: &str = "\u{E91B}";
 pub const COPY: &str = "\u{E916}";
 pub const DEBUG: &str = "\u{E90D}";
-pub const DECODERS: &str = "";
 pub const DOWNLOAD: &str = "\u{E912}";
 pub const FILE: &str = "\u{E91A}";
 pub const FOLDER: &str = "\u{E90A}";


src/main.rs
```diff
@@ -1081,7 +1081,8 @@ fn drawe(app: &mut App, gfx: &mut Graphics, plugins: &mut Plugins, state: &mut O
                     ui.with_layout(egui::Layout::right_to_left(Align::Center), |ui| {
                         drag_area(ui, state, app);
                         ui.add_space(15.);
-                        draw_hamburger_menu(ui, state, app);
+                        // This behaviour seems odd so i'm commenting it out. See #690 for details.
+                        //draw_hamburger_menu(ui, state, app);
```
woelper, line 1085: You can just remove the lines. I believe I wanted the menu in Zen mode in case someone got there by accident and needed a visual way out. I guess this can be removed if you folks like it better without.

woelper (approved):

src/main.rs
Stoppedpuma: I didn't even think about that reason and I do like the thought process surrounding it. Does it "fading out" and temporarily disappearing until the user moves their mouse seem better to you?

Issue #693, opened 2025-07-04
Add commands to move a user defined X%,Y% picture position to a user defined I%,J% window/screen position


eugenesvk:
**Is your feature request related to a problem? Please describe.**
This is most useful for ergonomically setting precise image positioning in image comparison, which was described in more details in https://github.com/woelper/oculante/issues/129.

**Describe the solution you'd like**
Add a set of custom commands that would allow positioning X%,Y% image coordinate to a I%,J% window or maybe even screen coordinates (since screen is the only guaranteed fixed coordinate system)

For example, if the center of an image is what you compare, you could add a command "move 50%,50% of an image to 50%,50% of the screen" and then have the perfect match (assuming matching zoom levels)

Of if you're comparing ↖corners, you could have "move 15%,15% of an image to 50%,50% of the screen"

I think ideally a command could allow you to 
  - input either (or both) values after it's invoked, 
  - but also you should be able to have custom commands with pre-defined coordinates

An additional ergonomic command would be to view/copy the current image % coordinates of the pixel that is positioned at, for example, the center of the screen. This way instead of repositioning both images you could copy position % from one and paste it into the repositioning command for the other

**Describe alternatives you've considered**
Careful mouse movements to achieve pixel-perfect alignment, but that's rather tedious

**Additional context**
This wouldn't work well non-proportionally sized images, so those use cases would need a different mechanism

eugenesvk:
No, these a very different. The other issue is about **Delta PX** movements, so it just extends the default Pan command to make it more powerful. Maybe it also makes sense to allow delta movements in % in addition to PX, but that's an extra flexibility there (interesting idea, but haven't thought about it).

This issue is about the ability to move to some **Absolute** position to make 2 pictures be positioned the same in 1 command (with delta panes it can take you a long while moving pictures around to make their positioning match even if the other FR is implemented as you'll eventually end up moving a picture by 1pixel and then switching to see if it matches, so rather inconvenient)

Stoppedpuma:
Thanks for the clarification! I'll have woelper look at this since I don't know how in-depth he would want this as I think the goal is still to keep Oculante in the middle of being mid weight and fully featured, just not to the level of say some KDE applications.

@woelper

PR #698, opened 2025-07-08
Add commands to move image position by percentage coordinates

nee-k:
This PR implements issue #693 by adding:
- A new command (Ctrl+P) to open a position dialog
- UI to specify image and window percentage coordinates
- Functions to calculate the necessary offsets
- Preset buttons for common positions (center, corners)
- A button to use the current cursor position

This allows users to precisely position images for comparison by aligning specific percentage points of the image to specific percentage points of the window.

Solved with [Zencoder](https://hubs.la/Q03vNcTr0)

### DIFF
--- a/src/appstate.rs
+++ b/src/appstate.rs
@@ -47,6 +47,32 @@ impl Message {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct PositionPercentages {
+    /// The X percentage of the image (0.0 to 100.0)
+    pub image_x: f32,
+    /// The Y percentage of the image (0.0 to 100.0)
+    pub image_y: f32,
+    /// The X percentage of the window (0.0 to 100.0)
+    pub window_x: f32,
+    /// The Y percentage of the window (0.0 to 100.0)
+    pub window_y: f32,
+    /// Whether the position dialog is open
+    pub dialog_open: bool,
+}
+
+impl Default for PositionPercentages {
+    fn default() -> Self {
+        Self {
+            image_x: 50.0,
+            image_y: 50.0,
+            window_x: 50.0,
+            window_y: 50.0,
+            dialog_open: false,
+        }
+    }
+}
+
 /// The state of the application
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -1036,6 +1043,9 @@ fn drawe(app: &mut App, gfx: &mut Graphics, plugins: &mut Plugins, state: &mut O
         if let Some(id) = state.filebrowser_id.take() {
             ctx.memory_mut(|w| w.open_popup(Id::new(&id)));
         }
+        
+        // Show position dialog if it's open
```
woelper, line 1047: I don't like that the function handles the visibility. You should do what your comment says and check if the dialog is open. I also don't like that application state is used for visibility. You should use egui memory for UI element visibility (ctx.memory for example).

src/appstate.rs
```diff
@@ -47,6 +47,32 @@ impl Message {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
```
woelper, line 50: Have you tested this? I don't see Serialize and Deserialze here. I am unsure how this would be saved/restored during restarts.

woelper (changes requested):

src/appstate.rs
Stoppedpuma: @woelper this PR doesn't work beyond the window opening and having sliders. This seems to be an "AI PR".

woelper: Thanks for your input! 
I am not generally against AI (assisted) PRs if they have been tested and follow standards - that said, I believe you are right.

I am inclined to close this and would like this to be opened as an issue first where we can debate how to solve this best (and if we should at all). I am also not 100% sure what problem is solved with this. Would you agree?

Stoppedpuma:
Hello! Oculante is currently going into maintenance mode to work on the next update which is a major rewrite. As such, this PR is being closed. More details can be found at #746. 

**We will only be accepting bug fixes until the rewrite is complete.**

PR #699, opened 2025-07-09
Fix most clippy lints

joshuamegnauth54:


joshuamegnauth54:
Okay! :grin: I'm using this build currently so if I find anything I'll fix it.

Stoppedpuma:
Thanks!

PR #704, opened 2025-07-15
Improve the UI of measure selection

Stoppedpuma:
Before (left) vs After (right) screenshot:

image


woelper:
This looks great! Thank you!

PR #709, opened 2025-07-23
Adjust hold key instruction colour

Stoppedpuma:
Makes the red a bit softer, matching the reference designs and red found in other areas like the warning toast.

Before:

image

After:

image


src/ui/settings_ui.rs
```diff
@@ -693,7 +693,7 @@ fn keybinding_ui(app: &mut App, state: &mut OculanteState, ui: &mut Ui) {
         ui.label_unselectable("While this is open, regular shortcuts will not work.");
         if no_keys_pressed {
             ui.label_unselectable(
-                egui::RichText::new("Please press & hold a key").color(Color32::RED),
+                egui::RichText::new("Please press & hold a key").color(Color32::from_rgb(255, 76, 76)),
```
woelper: I would recommend using the style, for example `style().visuals.selection.bg_fill` so it is consistent and centrally defined.

woelper (approved):

PR #719, opened 2025-07-28
feat: Add `{NUM}` to the `Window title` option

### LINKED ISSUE #413: {NUM} window title option
Shows the current number of images in the current directory. Example being 64/1024 meaning you are on image number 64 out of 1024 in the current directory.

brian6932:
closes #413


Stoppedpuma:
The position seems to be off by 1. It starts a 0 and will end a number short. For example, if you have 5 images, 1 will be 0 and the last image will be 4 instead of 5.

image

Stoppedpuma:
Thanks!

PR #720, opened 2025-07-29
feat: Fit window to image

brian6932:
This is a feature from qView I use, and really like.

- I haven't been able to get it to work [as fast and smoothly as qView though](https://github.com/jurplel/qView/blob/0ec246b78c310d5c842836a91566db24037c75c2/src/mainwindow.cpp#L542-L619). The performance of this PR doesn't feel good enough to be out of draft Imo. I actually can't get holding a folder direction (scrubbing) at all to really be as fast as qView (when on a fast drive; however, Oculante seems much faster when viewing things off a NAS, disregarding this PR), relative to qView's `Preloading = Extended`, and Oculante's `Number of images to cache = 1000`. 
- The scrubber, info, and header panels aren't offset/accounted for yet, should be doable with state though.
- I chose to center images at the center of the screen, instead of the center of the previous image, (which is what qView does), I don't think this really matters, as long as it's within bounds. This keeps it more in bounds than qView does.
- I don't know if there's a way to check whether a window's maximized in `notan`. qView doesn't resize maximized windows, and realistically this feature shouldn't either, but I haven't found a way to do that yet.
- Ideally `screen_size` would be determined before any windows were initialized, this isn't possible with `notan`. If it was, that'd remove the code in the `update` function, and avoid repeating the branch. I think `glutin` may support this, but I wasn't going to add a dependency for no reason here.
- There should be an event that gets spawned when you've moved the window from one monitor to another, currently `max_window_size` is determined by the launched monitor. Launching the program on a monitor's currently required to set this automatically.


Stoppedpuma:
> This feature's intended for people that use floating window managers, I don't see why it wouldn't work well across platforms, qView manages that just fine.

I say this because some window managers may have issues with this. For example, Hyprland is anything but stable and is filled with weird bugs that no other window manager has. A good example is right-click context menus just disappearing all together sometimes and requiring a complete restart of the application, or them just appearing in random spots around the screen.

Even on Sway, qView seems to like to resize the window towards the left side of the screen (or maybe this is intentional on qviews part?) instead of being in the centre. Nor can I get Oculante to run with this PR under Sway.

> Yea, one thought I had when writing this, was maybe storing the pre-computed window dimensions+positions queued per image, and then just apply them as a cached property, I'm unsure how much it'd help, but it's an idea I had regardless. Each time max_window_size and min_window_size change, you'd have to re-update the cache.

This would mean just expanding the current caching option we currently have. This works by caching to memory, not disk. Caching ahead would basically just cache the next x amount of images, caching the entire directory would be self explanatory. I'm assuming you're concern means that it would have to update if the user changes the fitting percentage?

brian6932:
> Am I understanding correctly that you just want the image to fit to the window at all times when a setting is active?

No, only when you first open the image. I want to be able to resize it after, intended to be used with `Fit image on window resize`, so you start out with the largest dimension, then can scale down. Basically this PR's a scuffed execution of qView's feature.

Stoppedpuma:
Hello! Oculante is currently going into maintenance mode to work on the next update which is a major rewrite. As such, this PR is being closed. More details can be found at #746. 

**We will only be accepting bug fixes until the rewrite is complete.**

PR #723, opened 2025-08-11
Fix compilation for new CMake (shaders)

joshuamegnauth54:
The GLSL feature of Notan fails to compile with newer versions of CMake. Notan also provides a shaderc feature for the same use case.

I added in a shaderc feature that's disabled by default as an alternative.


woelper:
I think it would be fair to try having `notan/glsl-to-spirv` as default instead of having a new `default_with_shaderc` flag. I forgot why `glsl-to-spirv` was default, but I suspect it was because of some os having issues. I'll start a PR and see if it builds on all platforms.

woelper:
The new PR is here, but NetBSD fails:
https://github.com/woelper/oculante/pull/724

PR #724, opened 2025-08-11
fix: make shaderc default for notan

woelper:


Stoppedpuma:
I've tested this on Arch Linux, Debian 13, and Windows 11. All seem to have no issue *running* Oculante with shaderc. I'll play around with this on Windows sometime tomorrow to make sure there's no windows specific issues. I'll also test FreeBSD.

Would you mind testing this on your ARM mac to make sure there's no issues on your side? The ARM Linux build would be good to test as well if you have the ability to do so.

Lastly, any idea on why NetBSD fails? Is it just incapable of running shaderc for one reason or another?

joshuamegnauth54:
Thanks for this! I recently updated my desktop but couldn't work on issues because Oculante wouldn't compile. :joy: 

I think the only issue with `shaderc` over `glsl-to-spirv` is that `shaderc` requires the program to be installed. `glsl-to-spirv` compiled glslang from source so it avoided that problem - that's likely why you had it as the default.

Should I close my old PR?

Stoppedpuma:
> there is a test for OSX which is arm, so every PR is always tested to compile.

I was meaning to check for runtime issues in-case there was anything that you previously ran into. Everything should be good to go though! Shaderc seems to work as usual on all places I've tested (Arch Linux, Debian 13, Windows 11, FreeBSD, MacOS Ventura (intel mac)). I that documenting the netbsd part is all that's left to do.

Stoppedpuma (approved):

Stoppedpuma:
This has been added to the README:

image

Stoppedpuma:
@joshuamegnauth54 Thanks for bringing this up! It was perfect timing as Debian 13 had just released.

I'll merge this now. Thanks again everyone!

Issue #710, opened 2025-07-23
HIF (HEIF) files from Cannon R10 HDR PQ mode do not display correctly


Stoppedpuma:
I can provide both the HIF and RAW privately for testing. The heif files from this camera seem to be pretty problematic as basically every viewer I have installed produces a different result. This photo was taken around 12 hours ago and has largely varying results. I think Gwenview is the closest to how this picture should actually appear if not slightly too bright / overexposed. At the bottom is the JPG output from the RAW to JPG conversion on the camera itself. It messes with the colours slightly, but it's still closest to GwenView.

RAW in darktable:

<img width="1777" height="1185" alt="Image" src="https://github.com/user-attachments/assets/018731d5-f157-475e-a160-3650cd3d4297" />

HIF in tev:

<img width="2161" height="1440" alt="Image" src="https://github.com/user-attachments/assets/82ff3638-4b64-49e4-b375-39ff8a065cdc" />

HIF: in GwenView

<img width="2172" height="1437" alt="Image" src="https://github.com/user-attachments/assets/0c61431a-cbfa-4043-aab6-cdb86ed3fe4a" />

HIF in Oculante:

<img width="2132" height="1437" alt="Image" src="https://github.com/user-attachments/assets/d994a69f-77dd-449e-8dba-265c7ed7b809" />

JPG in Oculante:

<img width="2122" height="1439" alt="Image" src="https://github.com/user-attachments/assets/125fffaa-5c9d-4305-98bf-b323e5e5c2b6" />

joshuamegnauth54:
It seems like Gwenview and darktable both use `libheif` like Oculante. I wonder if we need to change some settings on the HEIC context to fix how the image displays?

I'll do some research!

PR #731, opened 2025-09-10
Decode HEIC with correct color space

joshuamegnauth54:
Possibly closes: #710

I still have to test this so it's a draft for now.


Stoppedpuma:
These photos are a friends and I discovered this issue while they were viewing them on my laptop. As they are not mine, I'll ask for permission to send some of the originals for this PR. Do you have a preferred method I would be able to send a few of these privately to you?

joshuamegnauth54:
To clarify, some HEIC images worked but those specific images failed? If we can figure out the format, I can probably fix it without needing the images. If not, you could probably just email them (I think my email is in the commit).

For reference, I tested with these: https://heic.digital/samples/

src/image_loader.rs
```diff
@@ -173,34 +173,118 @@ pub fn open_image(
...
+                                .context("Failed creating RGBA image from HEIC")?,
+                        ),
+                        _ => unreachable!("Guarded by Rgb | Rgba"),
+                    }
                 }
-            }
-            let buf = image::ImageBuffer::from_vec(width as u32, height as u32, res)
-                .context("Can't create HEIC/HEIF ImageBuffer with given res")?;
-            let i = DynamicImage::ImageRgba8(buf);
+                Some(ColorSpace::YCbCr(chroma)) => {
```
awxkee, line 201: I'm not sure if this branch is correct one.

It's missing:

- Checking the image bit depth, and handling or rejecting unsupported depths. Typically, libheif supports 8-bit and 10-bit for HEIC.
- Accounting that YUV range is subject to change.
- Accounting for the YUV matrix, which may vary. For HEIC, it is usually BT.709 rather than BT.601. But it's needed to be queried from libheif.

src/image_loader.rs
```diff
@@ -173,34 +173,118 @@ pub fn open_image(
...
-                        data[step],
-                        data[step + 1],
-                        data[step + 2],
-                        data[step + 3],
-                    ]);
-                    step += 4;
+            let img = lib_heif.decode(&handle, ColorSpace::Undefined, None)?;
+
+            let i = match img.color_space() {
+                Some(ColorSpace::Rgb(chroma @ (RgbChroma::Rgb | RgbChroma::Rgba))) => {
```
awxkee, line 179: libheif can use 10 and 12 bits as well, not only 8 bit images.

src/image_loader.rs
```diff
@@ -173,34 +173,118 @@ pub fn open_image(
...
+                    )
+                }
+                Some(ColorSpace::Monochrome) => {
+                    let planes = img.planes();
+                    let y = planes.y.context("Grayscale image is missing luma")?;
+                    let buf = y.data.to_vec();
+                    let width = img.width();
+                    let height = img.height();
+
+                    let i = DynamicImage::ImageLuma8(
```
awxkee, line 266: Monochrome used in HEIC is YUV image with only Y plane, so it's required to handle YUV range and YUV matrix as well and image also could be 10 bit.

awxkee (commented):
This implementation is missing many cases that were previously handled internally by libheif.

It’s necessary to account for the following:

- 8/10 bit depth ( 12 bit depth HEIC's is possible but rare, AFAIK no opensource tools can encode 12 bit, so it's probably not required for HEIC's)
- YUV range changes
- YUV matrix changes
- GBR (identity matrix), which is technically YUV but slightly different and not especially rare.
- Monochrome is YUV 4:0:0 that should be handled appropriately 

There are also YCgCo images sometimes but as far as I know they are not handled out of the box in libheif so it should not be broken.

This won't help with HDR curves because they are different thing.

joshuamegnauth54:
@awxkee I recently realized that I didn't have to do any of this. `libheif` seems to convert the image on load so the original code was correct. With that said, do you have any ideas concerning the original issue? I would love to fix it.

@Stoppedpuma Yes, I got them and thank you! I have a slightly busy week, but I'll get back to work on this soon as I can.

awxkee:
@joshuamegnauth54 

Typical flow to handle HDR image without gainmaps:

1. Check CICP in image, if it is not present then exit early.
2. Check transfer function stored in CICP, if it is not PQ or HLG image is not an HDR, at least for CICP case.
3. As you determined that stored transfer function is HDR one ( PQ or HLG ) this is the time to start actually handle it.
4. You need to determine at first if driver + display supports requested color space + transfer function. This is system and driver dependent step. Though, I'm not familiar with framework here. Typically needed something like this in [Vulkan](https://registry.khronos.org/vulkan/specs/latest/man/html/VkColorSpaceKHR.html) or in [MacOS/iOS](https://developer.apple.com/documentation/coregraphics/cgcolorspace/itur_2100_pq?language=objc).
5. If HDR surface is available and you determined somehow that driver and display supports it, you just need create it, that's it.
6. If HDR surface is not available, then you need to perform a tonemapping. ITU-R has recommendation 2408, but this is recomendation only. Tonemapping itself is not strictly defined. It's kind of idea how to put something big into a something small. Thus, there are many possible approaches, but no single "exact" method. [Here](https://github.com/awxkee/gainforge/blob/4d1f27af780f8dcdadbb499fb85571fe88cd0f44/src/mappers.rs#L53) some tone mapper examples.
7. In most cases HDR tranfer function combined with Wide Gamut color spaces, if HDR surface is not available in most cases, wide gamut usually isn’t either. Therefore if you'll likely need along with tonemapping step convert Wide Gamut colorspace into sRGB, or some surface that you can create.

This is not actually as complicated as it may seems, but it does require some attention to details.

joshuamegnauth54:
Hmm, if this is dependent on the renderer then it may be best to wait until Oculante moves off Notan.

joshuamegnauth54:
I'll close this since it would need to be rewritten and because my initial approach was incorrect.

PR #738, opened 2025-10-18
fix(ui): Fix an issue where vertical resizing was not possible.

MugenTi:
- However, this is not a perfect solution. There is an issue where dragging the top, left, and right window borders causes the window height to gradually increase.

### DIFF
--- a/.github/workflows/check_osx.yml
+++ b/.github/workflows/check_osx.yml
@@ -21,9 +21,6 @@ jobs:
       - uses: ilammy/setup-nasm@v1
       - uses: Swatinem/rust-cache@v2
 
-      - name: Install shaderc
-        run: brew install shaderc
-
       # - name: Check without default features
       #   run: cargo check --no-default-features --features notan/shaderc
       #   if: matrix.os == 'macos-14'


Stoppedpuma:
Thanks for your work on this! I'm a bit unsure about this one however as I feel this gets in the way pretty easily. This wouldn't be as big of an issue if it didn't extent when resizing horizontally, but to me this feels worse than not being able to resize vertically.

MugenTi:
> Thanks for your work on this! I'm a bit unsure about this one however as I feel this gets in the way pretty easily. This wouldn't be as big of an issue if it didn't extent when resizing horizontally, but to me this feels worse than not being able to resize vertically.

I think that's a perfectly reasonable opinion.
I value the convenience of browsing files through a large window, but I doubt many people would tolerate such strange behavior.

MugenTi:
> glsl-to-spriv doesn't build on cmake >=3.5, the change to ShaderC was made recently because of this and will not be an issue anymore when we drop Notan.

I know absolutely nothing about shader compilers.
Note that I am building in a CMake version 4.1.2 environment.

MugenTi:
> An actions update PR can be found at #739. I'm going to probably end up rewriting the MacOS runner. To clarify, you are able to build glsl-to-spriv on both latest versions of windows and mac with their latest cmake versions?

Building glsl-to-spriv on Windows 11 is successful.
I do not own a Mac. Please refer to the test results on macOS below.
https://github.com/woelper/oculante/actions/runs/18654740449/job/53184747703?pr=738
Thank you for your work on #739.

Stoppedpuma:
Work regarding actions should be finished in that PR. I believe the test failures are just from Notan being outdated in one way or another at this point. Those should be resolved when we move away from Notan.

MugenTi:
While this doesn't solve the underlying issue, I've implemented a UI panel as an alternative. This has satisfied my personal needs.

File Browser Panel

Stoppedpuma:
That actually could be a good solution until we fix this (likely will come with the move away from Notan), I've had regular buttons on the floating windows since the beginning of when I started redesigning Oculante and later had the thought of using them for tiling. In your case, you've implemented a panel which I'm assuming works by splitting its area with the viewport so that selecting an image is always centred on the image portion of the viewport. Below are reference designs where you can see these buttons and how tiling a window would look.

Oculante_file-picker

image

This would follow the idea above, splitting the image viewport so that an image is centred based on the split when tiling is active. The following window buttons would be needed: Close (already have), Split Left, Split Right. These make up the three buttons in the designs above. Splitting up and down would be nice as well which we could use the left and right buttons for either with a combo like Shift+Click or just right clicking the buttons instead.

PR #739, opened 2025-10-22
Update actions, rewrite macOS runner, update dependencies

Stoppedpuma:
Updates runners to latest versions, appends -latest to some which aren't intentionally LTS.

Updated:

netbsd-vm 1.1.4 -> 1.2.0
ARM MacOS 14 -> latest
Intel MacOS 12 -> 15-intel
Ubuntu LTS 20.04 -> 24.04
Windows Server 2019 -> 2025
Dependency versions used by the runners

\+ cargo deps

Requesting Woelper to review to make sure I didn't overlook anything which could break anything special like intel mac tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,23 +95,21 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "android-activity"
-version = "0.6.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef6978589202a00cd7e118380c448a08b6ed394c3a8df3a430d0898e3a42d046"
+checksum = "0f2a1bb052857d5dd49572219344a7332b31b76405648eabac5bc68978251bcd"
 dependencies = [
  "android-properties",
  "bitflags 2.11.0",
  "cc",
- "cesu8",
- "jni 0.21.1",
- "jni-sys 0.3.0",
+ "jni",
  "libc",
  "log",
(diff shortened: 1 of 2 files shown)


Stoppedpuma:
@woelper This should be ready for review. The MacOS and NetBSD tests just refuse to build for some reason and randomly just broke one day. Likely more weirdness from Notan being out of date.

Stoppedpuma:
Moving this back to draft until I address NetBSD issues.

.github/workflows/check_osx.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       # - name: Check without default features
       #   run: cargo check --no-default-features --features notan/shaderc
-      #   if: matrix.os == 'macos-14'
+      #   if: matrix.os == 'macos-latest'
```
woelper: using "latest" is dangerous, as runners can change and the workflow breaks. I would recommend compiling on the oldest available runner as you will exclude users that have older macos versions.

.github/workflows/release.yml
```diff
@@ -58,76 +58,75 @@ jobs:
...
+
     - name: Install shaderc
       run: brew install shaderc
-      if: matrix.os == 'macos-14'
-
+      if: matrix.os == 'macos-15-intel' || matrix.os == 'macos-latest'
       
     - name: Install cargo deb
       run: cargo install cargo-deb
-      if: matrix.os == 'ubuntu-20.04'
```
woelper: in general, we might get complaints of the resulting binaries are used by users on older systems

woelper (approved):

PR #741, opened 2025-11-03
Option to auto scale smaller images to fit frame

joshuamegnauth54:
Currently, larger images automatically fit the window frame whereas smaller images are not scaled up. Smaller images take up only a fraction of the frame size. I usually have to zoom into smaller images to scale them up, but that becomes tedious for folders of many small images.

The default zoom option doesn't work as nice for mixed images sizes because images end up either zoomed in too much or not enough. What I want, and what this commit implements, is for images to always be scaled to the frame without any extra work.

I implemented this as an option that's disabled by default. It works for both smaller and larger images - in other words, it doesn't break images that are larger than the viewport. Those are still scaled correctly with this option enabled.


Stoppedpuma:
Thanks! This feature is definitely a must have!

PR #744, opened 2025-11-10
Pr/fix resize filter

MugenTi:
fix(resize): Overhaul resize filter logic for stable aspect ratio locking

- Completely rewrote the UI logic for the Resize filter to fix multiple issues.
- Aspect ratio is now managed as a state within egui's temporary data, ensuring that the ratio at the moment of locking is preserved and used consistently.
- Fixed a bug where the aspect ratio would fluctuate during drag input due to incorrect state handling.
- Fixed a bug where direct text input did not respect the aspect ratio lock.
- Real-time UI updates are now correctly synchronized, providing immediate feedback when changing dimensions with the lock enabled.

### DIFF
--- a/src/image_editing.rs
+++ b/src/image_editing.rs
@@ -138,6 +138,7 @@ pub enum ScaleFilter {
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
 
 pub struct ImgOpItem {
+    pub id: u64,
     pub active: bool,
     pub operation: ImageOperation,
 }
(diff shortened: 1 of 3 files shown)


Stoppedpuma:
The build seems to fail:

``` sh
error[E0609]: no field `new_image_loaded` on type `&mut oculante::appstate::OculanteState`
    --> src/main.rs:1050:18
     |
1050 |         if state.new_image_loaded {
     |                  ^^^^^^^^^^^^^^^^ unknown field
     |
     = note: available fields are: `image_geometry`, `compare_list`, `drag_enabled`, `reset_image`, `is_loaded` ... and 32 others

For more information about this error, try `rustc --explain E0609`.
error: could not compile `oculante` (bin "oculante") due to 1 previous error
```

Issue #711, opened 2025-07-27
Disable recent images


brian6932:
**Is your feature request related to a problem? Please describe.**
It's just a privacy thing, I don't like to have recent files enabled anywhere, I disable them in explorer too.

**Describe the solution you'd like**
Setting toggle.

**Additional context**
I'm coming from qView, and really like this image viewer, it's much faster, but missing a few things, so this will be one of many feature requests.

Stoppedpuma:
This will be customisable sometime soon.

brian6932:
Probably the former (as long as `0` disables collection).

Stoppedpuma:
> Should this be customizable by allowing the user to set how many images to remember or should it be a toggle that turns recent files on and off?

Being able to set a number value, 0 being off.

PR #755, opened 2025-11-17
Configurable max recent images

joshuamegnauth54:
Closes: #711

This also allows disabling storing recent images by setting the option to zero.


Stoppedpuma (approved):
Thanks!

Issue #766, opened 2025-12-18
Ability for file manager to open from directory of currently viewed image


Stoppedpuma:
Pretty big UX wise. This should be a keybind, shift + clicking or right clicking the icon on the bar should also probably do this.

PR #770, opened 2025-12-28
`last_open_directory` is incorrectly duplicated

joshuamegnauth54:
Related to #766. It turns out that the `last_open_directory` setting was saved in two different locations. It's saved in both the user's cache folder as well as in VolatileSettings. VolatileSettings is used in several places while the cache folder setting is only used in one place, so I assume this was an oversight.

As a plus, this simplifies the code in filebrowser.rs a bit too.


Stoppedpuma:
Great catch! I saw this at some point several months ago and I completely forgot about it. Thanks!

PR #778, opened 2026-01-16
Shift click to open browser in current img dir

joshuamegnauth54:
Closes: #766

Shift clicking the file browser button now opens the browser in the same directory as the currently viewed image.

### DIFF
--- a/res/oculante.desktop
+++ b/res/oculante.desktop
@@ -6,6 +6,6 @@ Keywords=image viewer;graphics;
 Exec=oculante %U
 Icon=oculante
 Terminal=false
-Categories=Graphics;
+Categories=Graphics;Viewer;
 Type=Application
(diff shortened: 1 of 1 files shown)


Stoppedpuma:
Thanks so much! This PR helps me out a ton! I've tested this just now on macOS and this seems to work great with the rfd feature! 

The only issue I can see that this PR has is this does not seem to work without the file_open (rfd) feature as this is something that should also work on Oculante's built-in file manager.

Stoppedpuma:
Thanks! Unfortunately I found (hopefully) one last edge case as I'm able to reproduce this still when making use of the compare feature.

To reproduce:

1. `cargo run /path/to/image`
2. shift + click to verify directory is originally correct
3. add to compare
4. Shift + click, add image from another dir
5. Shift + click yet again to check that dir is corect
6. add the second image to compare
7. Switch back to the first image and Shift click, the stale results show on either the first image, or the second one when you shift click.

Stoppedpuma (approved):

Issue #769, opened 2025-12-25
Revisit trackpad zooming and repositioning


Stoppedpuma:
Currently, zooming works by using two fingers specifically up or down. This is really weird and makes basically no sense (at least to me). Instead of this, we should follow the "pinch" design used by most software.

Moving around with two fingers, not pinching, should instead reposition the image.

Issue #760, opened 2025-12-03
Unable to run on Jay due to wayland-client <0.30.0


Stoppedpuma:
`wayland-client` must be at least version 0.30.0 to run under [Jay](https://github.com/mahkoh/jay)

https://github.com/mahkoh/jay/issues/679

PR #799, opened 2026-04-08
Oculante next

### LINKED ISSUE #746: Migration from Notan
Currently, Oculante is starting to fail to run and build on multiple systems, these in include the following:

- (Run) Ubuntu >=24.04
- (Run) Debian >=13 (GNOME)
- (Run) Arch Linux (GNOME)
- (Run) Anything using GNOME Wayland >=46 (Might affect earlier versions as well)

- (Build) NetBSD (May fail to run next release, fails to build master)
- (Build) NetBSD nasm >=3.0
- (Build) Linux with CMake >=4.0 (Currently known to affect NixOS build system)

There's a big issue which I've learned of from #743, NetBSD does not support Vulkan. This means that NetBSD may or may not be unable to build `notan/glsl-to-spirv` because of CMake 4.0+, nor can NetBSD use `notan/shaderc` since it requires Vulkan. This leads to the next section of this issue.

### Moving away from Notan:

Currently, there's been plans to move to Bevy, or to go without an existing engine, creating our own implementations. Bevy seems to be the preferred option and would solve all of our issues we are constantly hitting with Notan. We also wouldn't have to worry about it being abandoned any time soon since it's a large project with a source of income via sponsors. This is a good thing, Bevy is well backed, has desirable features, and makes development easier in the long term.

This seems perfect, right? Turns out, unfortunately not. @0323pin raised an issue at #743 regarding our current build issues and has tried to build a few Bevy examples and they will seemingly not run. This is our only major issue that we know of with a bevy implementation so far.

### What moving away from Notan accomplishes:

Notan has been our existing backend, it's done well I would say. So what's the issue with it? Notan is practically deprecated and is at the point where it's just causing more issues than it's worth. To keep this already long enough issue a bit shorter, this mainly consists of Notan specific bugs which will likely never be fixed in it's current state, parts of the codebase becoming a mess from Notan, preventing us from updating egui, and causing issues when building.

Moving away from Notan solves a bunch of problems and makes a bunch of improvements such as:

- A cleaner code base
- Less hacks to solve Notans shortcomings
- Being able to stay up to date with major dependencies like egui
- Solving build and runtime issues
- Fixes for Wayland
- Performance improvements
- Reducing build time

 And probably a thousand more that I can't begin to think of until we make an implementation

### Where this leaves us:

Oculante is probably the only project that does image viewing in an experience which I would describe as great or impressionable. I've introduced this project to several friends, all of which have fallen in love with Oculante. I would like this to run under as many operating systems as possible (within reason) for anyone to be able to use Oculante and experience this awesome program. The problem is, where do we go from here? Bevy seems to only run on the 4 major operating systems (Windows, macOS, Linux, FreeBSD). This prevents existing NetBSD users to continue use as well as from supporting other operating systems in the future.

So what do I think our options are? Well that's a bit of a handful. We could do any of the following as I see it:

- Attempt to get Bevy to run on NetBSD
- Create our own implementations
- Use / create an implementation specifically for NetBSD
- Drop NetBSD

If anyone reading this has anything to say on the matter, please feel free to leave a response!

### The next release (0.9.3)

As of writing this, It's been 10 months and a day since the last release of Oculante. To summarise, there's been a shortage of time to work on Oculante, and this big task isn't helping. A release of 0.9.2.1 could technically release at any point with a few changes and fixes. The next major release on the other hand should have these changes and will take a while to get things perfect and make a bunch of the changes we want to like removing old useless dependencies.

woelper:
This PR completely removes `notan` as a dependency. It will use egui exclusively for drawing operations. Please refer to testing.md for a todo list.


Stoppedpuma:
Hey! Sorry for getting to this so late, here's a quick review:

Info ui ignores theme and is always dark, dark colour is wrong, should be #191919.

ax

Modifying filter can shift image (this one is a little annoying to reproduce, move the image manually, then v to reset, then remove drag button to the right all the way, may take a few tries):

https://github.com/user-attachments/assets/bab37dc4-ce35-483a-b7b5-dd30c5853da3

Update position button in compare menu doesn't work

Modified and original buttons in edit menu don't work

Draw frame around image doesn't work

Info panel scroll bar is not in the correct location:

a

I think one thing we should handle while it's still pretty early on is updating to the latest egui version so that too many unnecessary changes aren't made. I've updated the testing.md with the things above as well as a few egui things I'm aware of.

woelper:
Cool, thanks!
I've found the info panel issue too and fixed it. There is a black bar remaining in the info panel though, which may be related to "Info panel scroll bar is not in the correct location".

Egui is already at 0.34 now.

Thanks for the rest, I'll add it to the testing.md!

Stoppedpuma:
A bit further of a review:

Floating window titles have this weird issue when the window is small and you press keys:

https://github.com/user-attachments/assets/b5666035-9c33-4091-8cb4-050c13046435

Changing system theme affects scale of parts of the UI, shifts it, restores on changing back to original theme:

https://github.com/user-attachments/assets/65376fc3-d5fa-43b4-b328-a3976d646ae1

APNG animations don't work, file stats update:

https://github.com/user-attachments/assets/61581bde-b5df-478b-9e38-b2e932c87891

Spacing on tiling text and slider is wrong, zoom preview also has a large gap (I just saw your note on this, I'll have to give this a bit of thought), image stats are also affected:

CleanShot 2026-04-10 at 21 05 38@2x

Resizing info UI to be smaller creates a black overlay in its place:

https://github.com/user-attachments/assets/45293d15-4011-472a-bb3a-871cb1a5b49d

Histogram x y text no longer shows:

https://github.com/user-attachments/assets/0896c56a-5457-458e-92cf-2058bfd6308a

Histogram has also changed, here's what it looks like in 0.9.2 on the same image:

CleanShot 2026-04-10 at 21 17 55@2x

woelper:
Thanks, I'll add these!
This is strange, I remember that exact apng working, but it does not animate either on my linux machine. Other apng files do work.

Stoppedpuma:
Another thing we can do cleanup wise is remove the egui-modal crate as egui now has built in support for modals.

woelper:
Great point. I think I ended up doing some custom stuff to achieve our style, I'll have a look what needs to change.

Overall I am quite happy so far, have not run into real deal breakers - but we also need to check performance later.

Stoppedpuma:
> Overall I am quite happy so far, have not run into real deal breakers

Out of curiosity, has there been anything notable so far?

Also regarding self_update, is the removal of the crate for us to use our own implementation insetad, or is the built-in updating feature just being outright removed?

woelper:
> > Overall I am quite happy so far, have not run into real deal breakers
> 
> Out of curiosity, has there been anything notable so far?

The main panic I had was with shaders and tiled rendering for large images, which worked. Egui seems to not support textures other than 32 bit rgba, so no native grayscale images to save memory or float texures, which is a bit sad. Maybe we can work around it later with something custom.


> Also regarding self_update, is the removal of the crate for us to use our own implementation insetad, or is the built-in updating feature just being outright removed?

I think it can be removed completely. It dragged in massive dependencies. Also, I saw problems on Windows and Mac as virus scanners and OS-level threat detection don't like binaries being swapped. So we'd do this only for Linux. I think at this point a regular distribution through native channels like package managers is best.

Stoppedpuma:
Makes sense, my only concern is just that Windows and Mac users generally don't use package managers. Would checking for a new release on launch be something for consideration? This would be a best of both worlds as it doesn't replace binaries, probably needs no new crates, lets us remove self_update, and still informs the user when there's an update available.

Reset view also doesn't seem take panels into consideration anymore:

https://github.com/user-attachments/assets/c82b6d93-ecfb-4af6-bdac-877454690d0b

The 3d2.png test doesn't render at all but image stats update, It doesn't generate a preview image in the file opener either.

woelper:
> Makes sense, my only concern is just that Windows and Mac users generally don't use package managers.
That is a good point. But at present, this behavior (at least in my tests) completely destroyed the app as OS-Level security would block the update and the resulting executable became 0 bytes and unusable. So if anyone used it, the app became broken.

With signing the app, we could release through brew or choco, which could make updates more seamless.

> Would checking for a new release on launch be something for consideration? This would be a best of both worlds as it doesn't replace binaries, probably needs no new crates, lets us remove self_update, and still informs the user when there's an update available.

If we do it, I would recommend we do a very basic check if there is a more recent version on github releases that is optional. But we'd reintroduce lots of otherwise unnecessary dependencies such as network, dns, https, semantic versioning etc.

> Reset view also doesn't seem take panels into consideration anymore:
> reset_view.mp4

Thanks, good find!

> The 3d2.png test doesn't render at all but image stats update, It doesn't generate a preview image in the file opener either.

I suspect the image does not decode at all with zune-png - I filed a new report here: https://github.com/etemesi254/zune-image/issues/372

Stoppedpuma:
>That is a good point. But at present, this behavior (at least in my tests) completely destroyed the app as OS-Level security would block the update and the resulting executable became 0 bytes and unusable. So if anyone used it, the app became broken.

It's definitely good to remove this if that's the case.

> If we do it, I would recommend we do a very basic check if there is a more recent version on github releases that is optional. But we'd reintroduce lots of otherwise unnecessary dependencies such as network, dns, https, semantic versioning etc.

Fine to drop it then in that case.

>With signing the app, we could release through brew or choco, which could make updates more seamless.

Signing is the only thing that's been preventing me from putting Oculante onto brew. As soon as we can get a signed version out there though I can get that PR out there!

App Store releases might interest you as well. Putting an app onto Apple's App Store is part of the Apple Developer Program which I believe you've already paid for to sign the application releases if I recall correctly. The Microsoft store is free for individuals to upload their apps.

https://developer.apple.com/support/compare-memberships/

a

Stoppedpuma:
Fit image on window resize is broken


https://github.com/user-attachments/assets/2ab64784-3d62-4070-bef0-1ae2422f2817

Stoppedpuma:
@woelper would you be able to take a glance at #769 to see about making this happen in this PR as well? This should be possible to do according to the multi touch example on egui.rs:


https://github.com/user-attachments/assets/24c7eba9-d0c1-49e8-b90f-5b0b664227a0

Stoppedpuma:
Another issue I just found, the keybinds section doesn't work properly with modifier keys such as shift or control.

https://github.com/user-attachments/assets/206af6a0-a632-4511-9ee3-31c0c69ba64d

woelper:
> @woelper would you be able to take a glance at #769 to see about making this happen in this PR as well? This should be possible to do according to the multi touch example on egui.rs:
> 
>  trackpad.mp4

Could be possible! I think this is mainly about detecting the gesture and getting a zoom increment. As soon as that is possible, pinch should work.

Stoppedpuma:
Another issue, you can change the zoom of the image when scrolling down menu items. This was done with a trackpad just in case it's not reproducible with a mouse.


https://github.com/user-attachments/assets/b21038ee-2857-4e23-847c-85a985cf93c7

Stoppedpuma:
@woelper I just wanted to ask if we're still going to be able to support HDR in the future? This question comes up because of the previous concerns (https://github.com/woelper/oculante/pull/799#issuecomment-4230146724). HDR has recently become huge factor for me and I would hate for us to lose out on supporting HDR.

Stoppedpuma:
Animated images appear to be broken again.

Stoppedpuma:
@woelper just as a heads up, Oculante 0.9.2 no longer opens on Sway due to the wl_shm error found in #760. This means that this error is now mainstream and affects a lot of users. I'm not sure if this affects KDE or GNOME but as far as I'm aware, Sway is pretty much the most popular wayland *window manager* and treated as a reference compositor in many cases.

This PR is of course unaffected, just current releases are.

I'm looking to help get a new update out as soon as possible so please let me know where my help would be most effective! I've also sent you an email a while back about a few ideas I've already had and started.

Stoppedpuma:
@panbergco I've implemented the copy path functionality that you were requesting. How it works in my implementation is in the hamburger menu, if you hold shift, the copy button changes to copy path. I've also implemented a global keybinding for copying the image path. By default the keybind is Ctrl+Shift+C Or CMD+Shift+C on macOS. 

https://github.com/user-attachments/assets/ef80ba10-f921-44c7-97d3-6cdb90e7f3f1

Stoppedpuma:
@woelper I've dropped zune-png in favour of image crates png implementation. This fixes our animated png issues that we were having as well as simplifies the loader code a bit.

If there was any reason in particular we used zune rather than image for png decoding, please let me know and I can revert this change if needed. I've also not done any benchmarks on this so I'm not sure if there's a performance impact or how big it would be.

woelper:
Awesome! If the samples load, all is great!

Stoppedpuma:
I've upgraded us to egui 35 but I've had to drop egui-notify in the process since it was outdated and is seemingly unmaintained. We now have our own toasts though!

https://github.com/user-attachments/assets/2818b545-788a-47ad-8f52-e27cc745c972

Stoppedpuma:
@woelper realistically, how much more work do you believe needs to be done on the glow renderer? As far as I'm aware, we can't render HDR images properly with glow so I think it might be a good option to just finish up your glow implementation and complete the list in the testing file so that we can create a release and work towards a wgpu implementation in the future? Using wgpu has benefits all around and the only downside that I can see is that NetBSD needs to play catch up in order to support wgpu.

PR #804, opened 2026-07-10
Add Copy file path shortcut and right-click context menu on image

panbergco:
## What

- **New shortcut `CopyFilePath`** (default <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>C</kbd>, <kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>C</kbd> on macOS) that copies the current image''s file path to the clipboard. It appears automatically in the Preferences keybinding editor, and existing users'' saved shortcut maps self-heal via the existing `key_pressed()` default-insertion mechanism.
- **Right-click context menu on the image area** offering *Copy image* and *Copy file path*. Uses the existing `styled_button` look & feel, honors the `pointer_over_ui` / `mouse_grab` guards, and closes on action, outside click, or <kbd>Esc</kbd>. Hover text shows the bound shortcuts.
- **"Copy path" entry in the hamburger menu** next to the existing *Copy*, following the same pattern.
- Small `clipboard_copy_text()` helper in `utils.rs` (arboard `set_text`).

## Why

Copying an image to the clipboard already exists as a menu entry and shortcut; copying the file's path is a frequent companion need (sharing paths in chats, terminals, scripts). A context menu on the image itself also makes both actions discoverable where users expect them.

## Notes

- No new dependencies.
- Ctrl+Shift+C does not collide with `Copy` (Ctrl+C), `CompareNext` (Shift+C) or the `C` channel key since `key_pressed()` matches on the exact number of held keys.
- Tested on Windows 11 (cargo check/build with default features, manual run).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -458,6 +458,15 @@ fn process_events(app: &mut App, state: &mut OculanteState, evt: Event) {
                 }
             }
 
+            if key_pressed(app, state, CopyFilePath) {
+                if let Some(p) = &state.current_path {
+                    clipboard_copy_text(&p.to_string_lossy());
+                    state.send_message_info("File path copied");
+                } else {
+                    state.send_message_warn("This image has no file path");
+                }
+            }
+
             if key_pressed(app, state, Paste) {
                 match clipboard_to_image() {
                     Ok(img) => {
@@ -1068,6 +1077,8 @@ fn drawe(app: &mut App, gfx: &mut Graphics, plugins: &mut Plugins, state: &mut O
         {
             toggle_fullscreen(app, state);
         }
+
+        image_context_menu(ctx, state);
 
         if state.new_image_loaded {
             ctx.memory_mut(|m| m.data.remove::<f64>(Id::new("resize_aspect_ratio")));
--- a/src/shortcuts.rs
+++ b/src/shortcuts.rs
@@ -39,6 +39,7 @@ pub enum InputEvent {
     LosslessRotateRight,
     LosslessRotateLeft,
     Copy,
+    CopyFilePath,
     Paste,
     Browse,
     Quit,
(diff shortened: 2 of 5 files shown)


Stoppedpuma:
First of all, thank you for the disclaimer that this PR is AI generated. There are a few things to take into consideration here but I will open by saying that If we do decide to merge this PR, it's not entirely up to me if we want to merge completely AI written code into the project as at the end of the day this is @woelper's project. From a quick glance at this PR, only a few changes are needed, but I do fear that merging AI code is going to encourage low effort AI PRs. Woelper has already stated that he is okay with AI **assisted** PRs of high quality, but I fear that even as little as having Claude in our contributors list could be a factor that turns away new and existing users as well as contributors.

That being said, let me actually get into a review:

To start, this PR is on the wrong branch. Master is currently not accepting new features and we have placed the project into maintenance mode. For this PR to even be considered, it would need to be refactored onto `oculante-next` as the target branch.

Next, this PR breaks the measure feature. There's an experimental measurement mode that's hidden behind a toggle in settings, enabling this shows the measuring option in the info menu. Here's a video showing this side by side:

https://github.com/user-attachments/assets/ba66b326-b555-41a4-bfe7-063f143cd186

Last, I would actually have just tucked this new option into the hamburger menu rather than creating a right click menu, either as a new button, or as an alternative behaviour when you hold shift.

Again, thank you for your interest in the project and for the AI disclaimer. For now, I'll close this PR as Master is in maintenance only mode and not currently accepting new features. In good news though, I will implement this myself in #799 with how I envision this should be implemented. In the future, please have any PRs you open centred around `oculante-next` until we merge this branch.

PR #807, opened 2026-08-31
Fix theme switch overwriting custom accent/background colors

Hai-Blahaj:
In settings, `apply_theme` was checking if an accent-/background-color is still a default by comparing it against the other theme's default value. So if you pick that exact color, it treats it as "not customized" and overwrites:

| Mode | Set accent to | changed to |
|---|---|---|
| Light | 255, 0, 75 | 0, 170, 255 |
| Dark | 0, 170, 255 | 255, 0, 75 |

Same for the background color, except it only triggers on theme switch and restarts, since its picker doesnt call `apply_theme` directly:

| Mode | Set background to | changed to |
|---|---|---|
| Light | 30, 30, 30 | 200, 200, 200 |
| Dark | 200, 200, 200 | 30, 30, 30 |

Fixed by tracking whether the user set the colors manually (`accent_color_is_custom`, `background_color_is_custom`).
Kept the old comparison next to it, so colors dont get reset on update.

Same bug present in oculante-next. Happy to PR that as well :)


Stoppedpuma:
Nice catch! I've rebased this onto `oculante-next` and I will merge this now. Thanks!

